PR #5, opened 2019-09-21
Migrate from Flow to TypeScript

mohoff:
Goal of this initial PR is to have a drop-in replacement for our existing `@brickblock/strong-config`. This means primarily that all flow-types are converted to Typescript typings. Additionally, typescript-specific project setup, CI setup, unit tests, dynamic type generation for Config object.

Donezo:

- [x] Setup Typescript project with tsconfig.json and eslint+prettier.
- [x] Migrate code base to Typescript
- [x] Refactored code into more composed utils
- [x] Generate Typescript types instead of Flowtypes for config object
- [x] Added test. Coverage above 80% 

Not in the scope of this MR:
- Make everything super nice
- Enduser-ready README, CONTRIBUTIONS, LICENSE
- line-perfect job definitions in `.gitlab-ci.yml`

These and more should be extracted into separate GitHub issues and addressed in future, hopepfully smaller, PRs


.eslintrc.js
```diff
@@ -0,0 +1,20 @@
+module.exports = {
```
chapati23, line 1: Any reason you haven't used our standard eslint config here?
```suggestion
module.exports = {
  extends: [
    '@brickblock/eslint-config-base-ts',
  ],
}
```

Admittedly it is a config custom-tailored to Brickblock's opinions. But in the end every eslint config is opinionated so why create the extra work of having to maintain another one for `strong-config`?

.gitignore
```diff
@@ -1,3 +1,5 @@
+lib
```
chapati23, line 1: curious, why do we need this? don't see a `lib` folder as part of this PR, is that generated by something else?

package.json
```diff
@@ -0,0 +1,70 @@
+{
+  "name": "strong-config-ts",
```
chapati23: this should be the npm package name to avoid issues

```suggestion
  "name": "@strong-config/node",
```

package.json
```diff
@@ -0,0 +1,70 @@
...
+  "version": "0.0.2",
+  "description": "SOPS-based secret management tool for NodeJS",
+  "main": "lib/index.js",
+  "types": "lib/index.d.ts",
+  "files": [
+    "lib"
+  ],
+  "repository": "git@github.com:strong-config/node.git",
+  "author": "mohoff <moritz.hoffmann@brickblock.io>",
```
chapati23: let's make this a team credit

```suggestion
  "author": "Brickblock Engineering <dev@brickblock.io>",
```

src/utils/hydrate-config.ts
```diff
@@ -0,0 +1,9 @@
+import R from 'ramda'
```
chapati23, line 1: not sure what the best practices for ramda are here, but isn't usually a selective import such as `import { compose, assoc, unary } from 'ramda'` easier to tree-shake and therefore leading to a smaller package?

.gitlab-ci.yml
```diff
@@ -0,0 +1,71 @@
...
+  - release
+
+# Fallback to support `includes`
+cache:
+  key: $CI_COMMIT_REF_SLUG
+  paths:
+    - node_modules/
+    - lib/
+
+.shared:
```
chapati23, line 16: interesting, never seen the dot notation before 🧐

what does it do?

chapati23:
Great, super excited to start working on this out in the open!

Before we move this PR out of WIP, i have the following requests:
- Add a little PR description on what this PR focusses on and therefore what reviewers should focus on. I assume a great chunk of the changed files are the result of `bbk sync`, so I'm not sure what I can ignore and where TypeScript-specific changes were introduced here
- Maybe also make explicit what's NOT in scope of this PR. E.g. if CI setup will be done in a separate PR that's totally fine (didn't see any checks in GitHub's UI, hence my impression that nothing was running)
- As this is our very first TypeScript MR of all times ever, it would be nice if you could add some quick thoughts on why you went with your chosen setup in case you deviated from any TypeScript defaults. E.g. is `lib` the standard build folder for TypeScript? If not, what made you chose it? Stuff like this.

.gitlab-ci.yml
```diff
@@ -0,0 +1,71 @@
...
+  only:
+    - master
+  except:
+    refs:
+      - schedules
+      - tags
+    variables:
+      - $CI_COMMIT_MESSAGE =~ /\[ci-release\]/
+  script:
+    - git config --global user.name "gitlab-bot"
```
crosby: @mohoff mind setting this and the email in the pipeline config (in GitLab) rather than hardcoding into this file. It's not too secret, but IMO we don't need to be sharing usernames and email addresses of our gitlab users here.

lint-staged.config.js
```diff
@@ -0,0 +1,9 @@
+module.exports = {
+  '*.ts': ['eslint --fix', 'git add'],
+  '*.json': ['jsonlint-cli', 'git add'],
+  '*.{yaml,yml}': ['yamllint', 'git add'],
+  '*.md': [
+    'markdownlint --ignore CHANGELOG.md --ignore node_modules --ignore .gitlab',
```
crosby: Why `.gitlab` here? Would be nice to also lint our `CHANGELOG.md`

scripts/healthcheck.ts
```diff
@@ -0,0 +1,75 @@
+/* eslint-disable @typescript-eslint/explicit-function-return-type */
```
crosby: @mohoff do we even need this file? In the husky file can we just do `yarn lint && yarn build && yarn todo`? Other than a cleaner output, I can't really think of any reason to wrap these commands in ora/shelljs.

src/load.ts
```diff
@@ -0,0 +1,64 @@
...
+
+const memoizedState: MemoizedState = {
+  didLoad: false,
+  config: null,
+}
+
+export const load = (
+  state: MemoizedState = memoizedState
+): Record<string, any> => {
+  if (state.didLoad) {
```
crosby: is `didLoad` even required? Can we just check if `state.config` is not `null`?

src/load.ts
```diff
@@ -0,0 +1,64 @@
...
+  const decrypted = sops.decryptToObject(
+    configFile.filePath,
+    configFile.contents
+  )
+
+  const config = hydrateConfig(process.env.RUNTIME_ENVIRONMENT)(decrypted)
+
+  const schemaFile = readSchemaFile('schema')
+
+  console.log('SCHEMA', schemaFile)
```
crosby: Would remove the `console.log` for this MR.

src/load.ts
```diff
@@ -0,0 +1,64 @@
...
+
+  if (schemaFile !== undefined) {
+    validateConfig(config, schemaFile.contents)
+    console.log('CONFIG and SCHEMA validated')
```
crosby: Same here.

src/utils/generate-type-from-schema.ts
```diff
@@ -0,0 +1,45 @@
...
+  filePath: string
+): Promise<void> => {
+  // TODO: remove `[k: string]: any;` in all generated interfaces. No option for this?
+  const baseTypes = await compileFromFile(filePath)
+
+  const schemaString = fs.readFileSync(filePath).toString()
+  const schemaObject = JSON.parse(schemaString)
+
+  // TODO: Make runtimeEnvironment configurable but keep a default
+  const configInterfaceAsString = `export interface Config extends ${pascalCase(
```
crosby, line 30: @mohoff the only place we use this library is here. A pure JS version:

`schemaObject.title.split(/[^a-zA-Z0-9]+/g).map(w => w[0].toUpperCase() + w.slice(1)).join('')`

This would allow us to remove a dep which is always nice 👍

src/utils/generate-type-from-schema.ts
```diff
@@ -0,0 +1,45 @@
...
+    schemaObject.title
+  )} {
+  runtimeEnvironment: string;
+}`
+  const exportedTypes = baseTypes.concat(configInterfaceAsString)
+
+  // TODO: Make filename configurable
+  fs.writeFileSync('strongconfig.d.ts', exportedTypes)
+}
+// Old logic
```
crosby: If new logic works, let's remove old logic comments. We have a record of this somewhere if we need to refer to it.

src/utils/substitute-with-env.ts
```diff
@@ -0,0 +1,20 @@
+import R from 'ramda'
+
+// TODO: Capture less strict /\${(.*)}/g so we can throw explicity when captured group violates /[a-zA-Z_]+[a-zA-Z0-9_]*/ (https://stackoverflow.com/a/2821201/3626075)
+const REGEXP_SUBSTITUTION_PATTERN = /\${(\w+)}/g
+
+export const substituteWithEnv = (stringContent: string): string =>
```
crosby, line 5: https://ramdajs.com/repl/?v=0.26.1#?const%20vars%20%3D%20%7B%0A%20%20SOME_ENV_VAR%3A%20%27test%27%0A%7D%0A%0Aconst%20config%20%3D%20%7B%0A%20%20a%3A%20%7B%0A%20%20%20%20b%3A%20%5B%0A%20%20%20%20%20%20%27SOME_ENV_VAR%27%0A%20%20%20%20%5D%0A%20%20%7D%0A%7D%0A%0A%0A%0Aconst%20substitute%20%3D%20curry%28map%28x%20%3D%3E%20%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20includes%28x%2C%20keys%28vars%29%29%20%3F%20%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20vars%5Bx%5D%20%3A%20%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20is%28Object%2C%20x%29%20%7C%7C%20is%28Array%2C%20x%29%20%3F%20%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20substitute%28x%29%20%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%3A%20x%0A%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%20%29%29%0A%0Asubstitute%28config%29%0A

scripts/healthcheck.ts
mohoff: Good point. I agree, a `yarn health` that calls `yarn lint && yarn build && yarn todo` seems good enough. Then we can keep the husky file as is.

src/load.ts
mohoff: Yeah works too. If we aim for explicitness, we can add a `isLoaded()` helper to our future `StrongConfig` class.

src/load.ts
mohoff: already gone

src/load.ts
mohoff: already gone

src/utils/generate-type-from-schema.ts
mohoff: On the other hand, the purpose of dependencies is also that you don't have to worry about these kind of snippets anymore. I feel like the problem of pascalCasing is widely known and has been solved in a popular library. Why re-implement a solution?

src/utils/generate-type-from-schema.ts
mohoff: already gone

lint-staged.config.js
mohoff: Copied that from an existing project. I remove `.gitlab` for now, however, `CHANGELOG.md` is autogenerated and full of markdownlint errors, as you can see in our existing projects.

mohoff: `.gitlab` could be created by the Gitlab runner though. We'll see in the next pipeline :)

src/utils/generate-type-from-schema.ts
crosby: In this case, installing `pascal-case` results in 7 more dependencies being added into the dep tree. It includes this dependency: https://www.npmjs.com/package/upper-case 😆 

Bloating the dep tree when the same can be achieved in a one-liner is an unfortuante side-effect of working within the JS ecosystem 😢

lint-staged.config.js
chapati23: yup, `CHANGELOG.md` is auto-generated by `standard-version`, therefore I don't see a benefit in linting it.

i think the `.gitlab` folder is just for issue and mr description gitlab templates. to change this we'd really need to change `project-templates` to differentiate between github and gitlab repos. not in scope for this MR IMO

package.json
```diff
@@ -0,0 +1,72 @@
+{
+  "name": "strong-config-ts",
+  "version": "0.0.1",
+  "description": "SOPS-based secret management tool for NodeJS",
```
chapati23: Nitty McNitterson: Official spelling is `Node.js` (according to on https://nodejs.org/en/)

scripts/healthcheck.ts
chapati23: Can't we just use the standard `healthcheck.js` from `project-templates` here? customising this for one repo only is configuration drift.

src/utils/generate-type-from-schema.ts
chapati23: it's an interesting discussion and depends on what we're optimizing for: DRY or performance?

As we're building a library here that's intended to be `yarn add`ed into other applications, we should probably be a bit more mindful of performance and our impact onto a user's `node_modules` folder and the dependency tree.

On the other hand, reinventing the wheel is also bad.

In this case, it's so little code that's affected that I would just inline the one-liner. Maybe with a comment a la `// Inlined for performance, inspired by <link-to-npm-package>`.

mohoff: Yeah bit philosophical but i'm sure we're removing here an obvious bloat, while there are many other modules which will still have these simple-stupid imports.

Anyway, i will update it.

src/utils/generate-type-from-schema.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  const configInterfaceAsString = `export interface Config extends ${pascalCase(
+    title
+  )} {
+  runtimeEnvironment: string;
+}
+`
+  const exportedTypes = baseTypes.concat(configInterfaceAsString)
+
+  fs.writeFileSync('strongconfig.d.ts', exportedTypes)
```
chapati23: nit: a bit off-brand. npm package is called `@strong-config` and github or is called `strong-config`, so i'd also make this filename `strong-config.d.ts`

src/utils/hydrate-config.ts
```diff
@@ -0,0 +1,10 @@
...
+import { substituteWithEnv } from './substitute-with-env'
+
+export const hydrateConfig = (runtimeEnvironment: string) =>
```
chapati23, line 4: Forgive my ramda illiteracy, what does this do exactly?

src/utils/read-file.ts
```diff
@@ -0,0 +1,103 @@
...
+    properties: {
+      runtimeEnvironment: string
+    }
+    required: string[]
+    sops: Record<string, any>
+  }
+  filePath: string
+}
+
+class NoFileError extends Error {
```
chapati23: interesting, is this the common way to create your own errors in typescript?

src/utils/read-file.ts
```diff
@@ -0,0 +1,103 @@
...
+  return parseConfig(files[0])
+}
+
+export const readSchemaFile = (filename = 'schema'): File | undefined => {
+  let schema
+
+  try {
+    schema = readConfigFile(filename)
+  } catch (error) {
+    // Support no schemas
```
chapati23: nit: reads a bit ambiguous. alternative suggestion: `// Defining a schema is optional. strong-config should still work without providing a schema.`

tsconfig.json
```diff
@@ -0,0 +1,24 @@
+{
```
chapati23, line 1: as this is the first time we use typescript and this config may define how we do it in other projects in the future: can you provide some more context on why you chose to configure it like this?

especially for cases where you deviated from whatever the 'standard' is (if there is one)

chapati23 (commented):
have to get used to this double-submission of review comments...

.eslintrc.js
mohoff: The goal is to go back to `@brickblock/eslint-config-base-ts`. To increase iteration speed and to have some room for experimentation, I used the plain rules here.

Let's backport to `@brickblock/eslint-config-base-ts` when we have a good setup.

.gitignore
mohoff: `tsconfig.json` defines an `outLib`, which is set to `./lib`. It's usually named the `dist` or `build` folder, but I've seen also `lib` for libraries. Felt like the more explicit choice.

src/utils/hydrate-config.ts
mohoff: After some research, I seems like tree-shaking is coupled with `webpack` and code for the browser. Also, https://github.com/ramda/ramda/issues/2355 suggests that there's no difference. Users importing `@strong-config/node`, which defines `ramda` as a dependency, will always install the full module I think.

src/utils/hydrate-config.ts
mohoff: 1. It replaces all occurrences of `${VAR}` with `process.env[VAR]`. The JSON-part is to facilitate the replacement so the problem is as simple as replacing captured regex groups in a string.
2. It adds the field `runtimeEnvironment: <runtimeEnvironment>` to the config object.

Overall it returns a function which is used as `hydrateConfig('development')(config)`

src/utils/read-file.ts
mohoff: Think so https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-2.html#example. Felt weird about that too

tsconfig.json
mohoff: TBH most of it is copied together from 'Getting started with TS' articles. These two seem required for libraries 
```
"declaration": true,
"declarationMap": true,
```

If you're already aware of better settings here, happy to discuss/update already. Otherwise, I'd update this as we go.

.gitlab-ci.yml
mohoff: done

lint-staged.config.js
mohoff: removed `.gitlab` but continue ignoring `CHANGELOG.md`

scripts/healthcheck.ts
mohoff: I'd ditch it because

1. it doesn't provide any value here. We can do `yarn health` as described above without the need of a `healthcheck.js`
2. IMO fine to deviate from our `project-templates`. Why is it relevant to enforce our `bbk init` setup here where 1. it's an open source project and 2. it's typescript which causes many deviations by itself already

src/load.ts
mohoff: Updated

src/utils/generate-type-from-schema.ts
mohoff: updated. I see it doesn't match the logic exactly. Made a comment in code. Eventually we have to update that in the future.

src/utils/substitute-with-env.ts
mohoff: It doesn't match with a pattern like `${...}` at all. I think it's important to _not_ replace the value in `{ someKey: 'USER' }` with `process.env[USER]`, which is what the above snippet does. Template literals are valuable here.

Furthermore, we use the same matching pattern as ES6 template literals and GitLab CI literals. Shouldn't be alien to users.

mohoff: Can we just leave it for now? I think it's too much thinking of performance optimization for the initial PR

src/utils/read-file.ts
mohoff: udpated

.eslintrc.js
chapati23: Interesting, what about `eslint-config-base-ts` slowed down your iteration speed exactly?

.gitignore
chapati23: `dist` and `build` is what i saw most of the time. do you have some examples of projects using `lib`? currently i'm gravitating more towards `dist` or `build`

.gitlab-ci.yml
mohoff: it means that it's not an actual job but a template to be used by jobs with `extends: .shared`

.npmignore
```diff
@@ -0,0 +1,8 @@
+src
```
chapati23: IIRC you don't need an `.npmignore` file when using the `files` array in `package.json`. from what i remember, `.npmignore` is the blacklist approach of `Publish every file EXCEPT these`, whereas when you define a `files` array in `package.json` it's the whitelist approach of `Publish ONLY these explicitly listed files and folders`

.gitlab-ci.yml
```diff
@@ -0,0 +1,71 @@
...
+  paths:
+    - node_modules/
+    - lib/
+
+.shared:
+  artifacts:
+    paths:
+      - lib/
+      - package.json
+      - CHANGELOG.md
```
chapati23, line 21: wanted to first write that we should also include `README.md` here, but then realised i don't fully understand what the artifacts are being used for.

can you comment on what these artifacts are being used for?

.gitignore
mohoff: for example https://basarat.gitbooks.io/typescript/docs/quick/library.html , besides the official docs the go-to guide

mohoff: and there are more articles suggesting the same https://www.google.com/search?ei=N-KNXaDMOMK4kwXXupjwBQ&q=lib+folder+typescript+library&oq=lib+folder+typescript+library&gs_l=psy-ab.3..0i71l8.0.0..5849...0.2..0.0.0.......0......gws-wiz.Nmxfz6dBQs0&ved=0ahUKEwjgw8fI4_DkAhVC3KQKHVcdBl4Q4dUDCAs&uact=5

As i said, `lib/` seemed good because it's explicit

.eslintrc.js
mohoff: Figuring out `eslint-config-base-ts` itself, seeing what needs overwrites and what not, ...

I know that rules can be overwritten if smth is off, but then I preferred no extend rather than a off one. If you care about this, let me know and I align the eslint setup in this MR

package.json
mohoff: really

src/utils/hydrate-config.ts
chapati23: i see, thanks for the link, [this was the relevant answer for me](https://github.com/ramda/ramda/issues/2355#issuecomment-341513380). would still be curious if the result is the same two years later.

resolving this issue for now

src/utils/hydrate-config.ts
chapati23: thanks

src/utils/substitute-with-env.ts
chapati23: to be honest, i find the existing code very easy to read and not overly long. ramdaifying feels a bit like developer masturbation 🍆🙈

tsconfig.json
chapati23: i'm not, was just wondering whether there were any conscious decisions you took here that deviate from a standard because of some tricks you've learned about TypeScript :)

chapati23 (changes requested):
still figuring out what useful comment to write here when submitting a review :D

.eslintrc.js
chapati23: Ok, doesn't need to be in this PR, but I do think it's best to use our full config sooner rather than later.

Feel free to resolve this after you create a GH issue to migrate to our config in one of the next PRs

package.json
chapati23: yeah, i know. this will show up when searching our package on NPM and other places so i do believe the attention to detail demonstrates the amount of care that went into a product or package.

pretty please 🦄

.gitignore
chapati23: ok, convinced, thx for the links 👍

.gitlab-ci.yml
chapati23: TIL, thx

src/utils/read-file.ts
chapati23: thx. 

sidenote: gitlab does a better job here of auto-updating comments with links to new commits

scripts/healthcheck.ts
chapati23: it just makes a lot of things easier to stick to a standard. this won't be our last open source project. and if we want to use typescript internally, we will create a healthcheck for it anyways. 

regarding ditching this file: we used to do just `yarn lint:bla && yarn lint:blub` but the sentiment then was that it was a very noisy task to run locally (where it's primarily run) and made it difficult to see what went wrong and where. which was the reason we came up with the clean formatting in the first place.

src/utils/read-file.ts
chapati23: ok

.gitlab-ci.yml
mohoff: artifacts here are files that can be downloaded by the upcoming jobs in a pipeline.

This is not superclean here since the `include`s don't use the `extends` syntax. The set of artificats is preliminary but I'm sure that we don't need a `README` as artifact right now.

chapati23: Ok, I know what the general purpose of artifacts is (to share data between pipelines). What are we actually _using_ artifacts here for? What’s the reason a subsequent pipeline needs access to a previous pipeline’s output in this repo?

mohoff: ok you sounded like you need a primer on artifacts.

You're completely right, these lines are not final. As you know, the general purpose is to not do redundant computations but rather cache outputs via `cache` and `artifact`. Before we now excerpt the `.gitlab-ci.yml` line by line, i'd encourage us to be happy about a green pipeline and make it nice in another MR. I will also figure out a good `extend` structure for our internal pipelines in this sprint.

You'll hear back regarding this in another MR

.npmignore
mohoff: Thanks for the hint! I'll update

.eslintrc.js
mohoff: New issue: https://github.com/strong-config/node/issues/12

Are you interested in forming the ruleset for this?

scripts/healthcheck.ts
mohoff: Do you know why we have JS and shell healthchecks? The shell-based one with `chronic` seems useful and I agree with here with your point of better output. What's the purpose of `healthcheck.js` though?

.gitlab-ci.yml
mohoff: Created https://github.com/strong-config/node/issues/13

PaddyMc:
Should we add a CONTRIBUTING.md file? It may help attract open-source devs.

[Reasoning](https://mozillascience.github.io/working-open-workshop/contributing/)
[Example](https://gist.github.com/PurpleBooth/b24679402957c63ec426)

Maybe not for this PR tho, wdyt?

README.md
```diff
@@ -0,0 +1,13 @@
...
+
+Secret management for NodeJS. Based on [SOPS](https://github.com/mozilla/sops).
+
+[![pipeline status](https://git.brickblock.sh/devops/strong-config-ts-mirror/badges/master/pipeline.svg)](https://git.brickblock.sh/devops/strong-config-ts-mirror/commits/master)
+
+[![coverage report](https://git.brickblock.sh/devops/strong-config-ts-mirror/badges/master/coverage.svg)](https://git.brickblock.sh/devops/strong-config-ts-mirror/commits/master)
+
+## How-to
+
+Set the environment variable `RUNTIME_ENVIRONMENT`
```
PaddyMc, line 11: Can we improve this, so it's easier for someone to clone the project and start developing?

1. Prerequisites e.g sops must be installed => google kms => account on google cloud with permissions to x, y, z 
2. How to start developing 
3. How to test
4. How to use

wdyt?

chapati23:
> Should we add a CONTRIBUTING.md file? It may help attract open-source devs.
> 
> [Reasoning](https://mozillascience.github.io/working-open-workshop/contributing/)
> [Example](https://gist.github.com/PurpleBooth/b24679402957c63ec426)
> 
> Maybe not for this PR tho, wdyt?

Great idea, can you create a new ticket for this @PaddyMc ?

README.md
chapati23: Agree but I'd say this is a separate MR.

(@mohoff these are the types of comments one typically avoids with a clearer MR description)

chapati23: Should probably also warrant its own ticket, our README needs to be ace. If we want people to adopt this, the README is probably their first touchpoint.

.npmignore
mohoff: updated

README.md
mohoff: Definitely will do that!

But not related to `Migrate from Flow to Typescript`. Would appreciate a ticket @PaddyMc

PaddyMc: Ace, just created two issues :)

scripts/healthcheck.ts
chapati23: I believe the main reason for switching to a JS-based approach was that you don't have any dependency issues. If it was a bash script, every developer needs to locally have `chronic` available AND we'd need to include the `chronic` binary in our CI image which comes with its own overhead.

So we said, hey why don't we make this just another `devDependency` same as all other dev tooling.

.eslintrc.js
chapati23: Thx, will resolve.

My suggestion for the initial ruleset is effectively [eslint-config-base-ts](https://git.brickblock.sh/core/eslint-config-base-ts) which simply uses [@typescript-eslint/recommended](https://github.com/typescript-eslint/typescript-eslint/blob/67537b863a8a98383796abb404224fdfc1c7b996/packages/eslint-plugin/src/configs/recommended.json)

If we want to customise the TypeScript rules a bit more, I think you'd probably be more qualified to do this by now as I haven't really done a lot of TypeScript development in the last 2 years.

.gitlab-ci.yml
chapati23: alright, let's discuss in more detail in the ticket.

scripts/healthcheck.ts
mohoff: Thanks.

No strong opinion. From my side:
- `chronic` won't be required for users, just for devs. This is us and eventual external contributors. Seems fine, especially with making this part of `CONTRIBUTION.md`.
- `chronic` is not required in any images because we're never running `yarn health` in CI - only locally.
- Having a `healthcheck.js` in a typescript project seems off. Would favor `healthcheck.ts` that is called with `ts-node`.

Please let me know your choice.

chapati23: I think the amount of code we're talking about here is fairly minimal, so IMO paying a few more lines of code for 
1) Getting rid of `chronic` as an external dependency
2) Keeping everything in JS

is a good tradeoff.

If you want to refactor it to TypeScript, that's fine with me. However, a simple `node scripts/healthcheck.js` will then no longer suffice and we need some kind of typescript binary. (Does something like `ts-node` come bundled with the `typescript` npm package?)

mohoff: done

chapati23 (approved):

PR #16, opened 2019-10-04
Remove index signatures generated by json-schema-to-typescript

### LINKED ISSUE #6: Improve generated Typescript interfaces
Currently, the compiled types are written into a `strongconfig.d.ts`. It also exports an interface representing the top-level `Config` type.

1. Investigate,
    -  if there's a better util to generate types. 
    - if it makes more sense to generate `type`s instead of `interface`s
2. Improve type generation with:
    - Remove `[k: string]: any` of all generated interfaces
    - Implement check that `title` of a json-schema definition can not be named `Config` as it's reserved for the top-level interface we append
    - Improve header: Replace warning. Is `tslint:disable` needed?

mohoff:
Remove index signatures that are auto-generated with `json-schema-to-typescript`.

This keeps all generated types exact so they match with config. For example, if the config defines

```yaml
logger:
   level: ...
   name: ...
```

and exports `interface Logger`, then it shouldn't be possible to do:

```js
const logger: Logger = {
   level: ...
   name: ....
   fieldThatIsNotInConfig: ....  // should cause error
}
```

We want to only allow:

```js
const logger: Logger = {
   level: ...
   name: ....
}
```

Fixes #6

### DIFF
--- a/src/utils/generate-type-from-schema.test.ts
+++ b/src/utils/generate-type-from-schema.test.ts
@@ -27,6 +27,15 @@ const mockedSchemaStringWithInvalidTitle = `
     "description": "This is a description"
   }
 `
+const mockedType = `export interface TheServerSchema {
+  port: ThePort;
+  [k: string]: any;
+}
+`
+const mockedTypeWithoutIndexSignature = `export interface TheServerSchema {
+  port: ThePort;
+}
+`
 
 jest.mock('fs')
 jest.mock('json-schema-to-typescript')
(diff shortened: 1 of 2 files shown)


chapati23 (changes requested):
After reading a little bit into the topic, I don't think we should do this.
Check out [this section](https://json-schema.org/understanding-json-schema/reference/object.html#properties) in the `json-schema` docs. Quoting the relevant parts:

> By default, providing additional properties is valid.
> The `additionalProperties` keyword is used to control the handling of extra stuff, that is, properties whose names are not listed in the properties keyword. By default any additional properties are allowed.
The `additionalProperties` keyword may be either a boolean or an object. If additionalProperties is a boolean and set to false, no additional properties will be allowed.

So it seems we've just not been using the full power of JSON schema so far. Doing this would get rid of `[k: string]: any` without any additional code:

```json
{
  "title": "Example Schema",
  "type": "object",
  "properties": {
    "firstName": {
      "type": "string",
    },
    "lastName": {
      "type": "string"
    },
  },
  "additionalProperties": false,
  "required": ["firstName", "lastName"]
}
```

mohoff:
Ahh I see. Yeah, then we haven't been using it in the correct way.

I agree with your point.

PR #17, opened 2019-10-04
Chore/better types

mohoff:
This PR introduces properly typed config objects.

Donezo:
- [x] Accurate types for intermediate and final config objects
- [x] small improvements to the eslint rules have been made (soon to be backported to `@brickblock/eslint-config-base-ts`)


prettier.config.js
```diff
@@ -2,4 +2,6 @@ module.exports = {
   semi: false,
   singleQuote: true,
   trailingComma: 'es5',
+  bracketSpacing: true,
```
chapati23: this is `true` by default. can safely be omitted

chapati23: Check https://prettier.io/docs/en/options.html#bracket-spacing

prettier.config.js
```diff
@@ -2,4 +2,6 @@ module.exports = {
...
+  arrowParens: 'avoid',
```
chapati23: This is also already the default setting and can safely be omitted: https://prettier.io/docs/en/options.html#arrow-function-parentheses

src/config.d.ts
```diff
@@ -0,0 +1,27 @@
+// These types be improved once recursive type references are released (https://github.com/microsoft/TypeScript/pull/33050)
```
chapati23: This looks like a todo, so please add `// TODO:` here ✅

src/config.d.ts
```diff
@@ -0,0 +1,27 @@
...
+
+// eslint-disable-next-line @typescript-eslint/no-empty-interface
+interface JSONArray extends Array<JSONValue> {}
+type JSONValue = string | number | boolean | JSONObject | JSONArray
```
chapati23, line 5: This feels a bit boilerplaty that we have to define this ourselves. I assume you wanted a type that is `anything BUT null|undefined`.

Is there no typescript-native way to do this?

src/config.d.ts
```diff
@@ -0,0 +1,27 @@
...
+  kms: unknown | null
+  gcp_kms: unknown | null
+  azure_kv: unknown | null
+  pgp: unknown | null
+  lastmodified: string
+  mac: string
+  version: string
+  encrypted_suffix?: string | null
+}
+// For explicitness
```
chapati23: i like comments, but this one doesn't tell me much. explicitness about or for what?

src/load.test.ts
```diff
@@ -71,7 +75,7 @@ describe('load behaves as expected', () => {
     process.env = OLD_ENV
   })
 
-  it('returns the memoized state if didLoad is true', () => {
+  it('returns the memoized state loaded config is not nil', () => {
```
chapati23: grammar is off here. are you missing an `if`?

src/load.ts
```diff
@@ -18,16 +17,19 @@ export const clearMemoizedConfig = (): void => {
...
-    memoizedConfig = passedConfig
-    return passedConfig
-  }
-
+): HydratedConfig => {
   if (R.isNil(process.env.RUNTIME_ENVIRONMENT)) {
     throw new Error('process.env.RUNTIME_ENVIRONMENT must be defined.')
   }
 
+  if (passedConfig !== undefined) {
```
chapati23: Could use the short form here:

```suggestion
  if (passedConfig) {
```

src/utils/hydrate-config.ts
```diff
@@ -1,7 +1,13 @@
 import R from 'ramda'
 import { substituteWithEnv } from './substitute-with-env'
 
-export const hydrateConfig = (runtimeEnvironment: string) =>
+export type InnerHydrateFunction = (
```
chapati23, line 4: the name confuses me a bit, is there a more expressive one maybe? what is this function doing exactly?

src/utils/read-file.ts
```diff
@@ -9,10 +9,7 @@ export enum FileExtension {
   YML = 'yml',
 }
 type File = {
-  contents: {
-    sops?: Record<string, any>
-    [key: string]: any
-  }
+  contents: EncryptedConfig | Schema
```
chapati23, line 12: 👏

src/load.ts
chapati23: or, for consistency with the other nil checks:

```suggestion
  if (!R.isNil(passedConfig)) {
```

chapati23 (changes requested):
looks much nicer, a few smaller comments

src/config.d.ts
mohoff: Hm it's not the same since there are many more types that are not valid JSONValues: `Symbol`, `Function`, `Enum`, any class, maybe more.

So it's correct that a valid JSONValue is defined as shown in L5. Also safer bc more defensive.

src/config.d.ts
mohoff: it's an alias, not a distinct type.

src/config.d.ts
chapati23: ok, just feels like something as basic as JSON types this should be baked in to typescript. but if there's no better way to do this: 👍

src/utils/hydrate-config.ts
mohoff: Open for suggestions.

`hydrateConfig` is curried. It does the same thing as before, which i described in the first PR already (https://github.com/strong-config/node/pull/5#discussion_r328724018). Just with types so it should be clearer than before.

src/load.ts
mohoff: addressed

src/load.test.ts
mohoff: fixed

src/config.d.ts
mohoff: updated

prettier.config.js
mohoff: removed

prettier.config.js
mohoff: removed

chapati23 (approved):

PR #18, opened 2019-10-07
Feat/validate

### LINKED ISSUE #10: Expose config validation against schema
Currently, the exposed `validate()` method is just a placeholder and should be replaced with actual config validation.

Possible functionality of `validate()`:

- It should validate all config files in the config directory
- It should warn or fail when there is no schema file found
- It should provide validation errors on a per-file basis
- It can accept a config file path as string (or array of strings?) which narrows the validation down to one config instead of all.

mohoff:
### Donezo:

- [x] Expose a `validate()` function besides existing `load()`
- [x] Refactor io-related code to make it more generic, re-usable and easier to test

The `validate()` functions returns `true` if the validation(s) pass, and throws an Error otherwise. It can be used as part of tests, and/or CI to validate configs without returning them. The function signature is:

```js
// x = path to schema file
// y = path to config folder or a config file
function validate = (<x>, <y>, [...<y>]) => true | Error

// Example calls
validate('config/schema.json', 'config')
validate('config/schema.json', '../config')
validate('config/schema.json', '../config', '/path/to/other/config/file.yaml')
```

In words: Firstly, it requires a `schema.json` to validate against. Secondly, it accepts a list of config paths (can be dirs or files), which are passed as variadic parameters.

fixes https://github.com/strong-config/node/issues/10

### How to test

1. Checkout this branch, Checkout another project like api-gateway locally (master)
2. `yarn build` strong-config
3. in api-gateway: `yarn remove @brickblock/strong-config`, `yarn add <path to local 
@strong-config/node>`
4. In `api-gateway/src/index.js`, replace occurrences of `@brickblock/strong-config` with `@strong-config/node`. In the same file, add

```js
const validationResult = require('@strong-config/node').validate(
  'config/schema.json',
  'config/development.yml'
)
console.log(validationResult)
```

5. When you `yarn start` the api-gateway, it should print `true` to the console. If you play around with config files or the schema, it will throw instead.

### DIFF
--- a/src/load.test.ts
+++ b/src/load.test.ts
@@ -61,7 +61,7 @@ const alreadyLoadedConfig = {
   runtimeEnvironment: RUNTIME_ENVIRONMENT,
 }
 
-describe('load behaves as expected', () => {
+describe('load()', () => {
   const OLD_ENV = process.env
 
   beforeEach(() => {
@@ -92,7 +92,10 @@ describe('load behaves as expected', () => {
   it('reads the config based on process.env.RUNTIME_ENVIRONMENT', () => {
     load()
 
-    expect(mockedReadConfigFile).toHaveBeenCalledWith(RUNTIME_ENVIRONMENT)
+    expect(mockedReadConfigFile).toHaveBeenCalledWith(
+      expect.any(String),
+      RUNTIME_ENVIRONMENT
+    )
   })
 
   it('decrypts the config with SOPS', () => {
--- a/src/utils/find-files.test.ts
+++ b/src/utils/find-files.test.ts
@@ -1,0 +1,103 @@
+jest.mock('glob')
+jest.mock('path')
+jest.mock('./read-file')
+import glob from 'glob'
+import path from 'path'
+import { getFileExtensionPattern } from './read-file'
+const mockedGlob = glob as jest.Mocked<typeof glob>
+const mockedPath = path as jest.Mocked<typeof path>
+const mockedGetFileExtensionPattern = getFileExtensionPattern as jest.MockedFunction<
+  typeof getFileExtensionPattern
+>
+
+const mockedFileNames = ['/dev/config/development.yml', 'schema.json']
+const mockedResolvedPath = '/dev'
+mockedGlob.sync = jest.fn().mockReturnValue(mockedFileNames)
+mockedPath.resolve = jest.fn().mockReturnValue(mockedResolvedPath)
+mockedGetFileExtensionPattern.mockReturnValue('json')
+
+import { findConfigFiles, findFiles, isSchema } from './find-files'
+
+describe('findFiles()', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('invokes glob.sync', () => {
+    findFiles('some/path')
+
+    expect(mockedGlob.sync).toHaveBeenCalledTimes(1)
+  })
+
+  it('returns result of glob.sync', () => {
+    expect(findFiles('some/path')).toEqual(mockedFileNames)
+  })
+
+  it('invokes glob.sync with correct default glob', () => {
+    findFiles('some/path')
+
+    expect(mockedGlob.sync).toHaveBeenCalledWith('**/*.*', expect.any(Object))
+  })
+
+  it('builds non-default glob correctly', () => {
+    findFiles('some/path', 'schema', 'json')
+
+    expect(mockedGlob.sync).toHaveBeenCalledWith(
+      'schema.json',
+      expect.any(Object)
+    )
+  })
+
+  it('invokes glob.sync with correct options', () => {
+    findFiles('some/path')
+
+    expect(mockedGlob.sync).toHaveBeenCalledWith(expect.any(String), {
+      cwd: mockedResolvedPath,
+      absolute: true,
+    })
+  })
+})
+
+describe('isSchema()', () => {
+  it('returns false for paths where the filename does not start with schema', () => {
+    expect(isSchema('/dev/config/development.yaml')).toBe(false)
+  })
+
+  it('returns true for paths where the filename starts withs schema', () => {
+    expect(isSchema('/dev/config/schema.json')).toBe(true)
+  })
+})
+
+describe('findConfigFiles()', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('invokes glob.sync', () => {
+    findConfigFiles('some/path')
+
+    expect(mockedGlob.sync).toHaveBeenCalledTimes(1)
+  })
+
+  it('returns result of glob.sync but filters out files named schema', () => {
+    expect(findConfigFiles('some/path')).toEqual([mockedFileNames[0]])
+  })
+
+  it('invokes glob.sync with correct default glob', () => {
+    findConfigFiles('some/path')
+
+    expect(mockedGlob.sync).toHaveBeenCalledWith(
+      '**/*.json',
+      expect.any(Object)
+    )
+  })
+
+  it('builds glob correctly when fileName is passed', () => {
+    findConfigFiles('some/path', 'schema')
+
+    expect(mockedGlob.sync).toHaveBeenCalledWith(
+      'schema.json',
+      expect.any(Object)
+    )
+  })
+})
--- a/src/utils/find-files.ts
+++ b/src/utils/find-files.ts
@@ -1,0 +1,28 @@
+import R from 'ramda'
+import path from 'path'
+import glob from 'glob'
+
+import { getFileExtensionPattern } from './read-file'
+
+export const isSchema = (filePath: string): boolean =>
+  R.compose<string, string[], string, boolean>(
+    fileName => fileName.startsWith('schema'),
+    R.last,
+    R.split('/')
+  )(filePath)
+
+export const findFiles = (
+  basePath: string,
+  globFileName = '**/*',
+  globFileExtension = '*'
+): string[] => {
+  const globPattern = `${globFileName}.${globFileExtension}`
+
+  return glob.sync(globPattern, { cwd: path.resolve(basePath), absolute: true })
+}
+
+export const findConfigFiles = (
+  basePath: string,
+  fileName?: string
+): string[] =>
+  R.reject(isSchema, findFiles(basePath, fileName, getFileExtensionPattern()))
--- a/src/utils/get-file-from-path.test.ts
+++ b/src/utils/get-file-from-path.test.ts
@@ -1,0 +1,94 @@
+jest.mock('fs')
+jest.mock('js-yaml')
+
+import fs from 'fs'
+import yaml from 'js-yaml'
+
+const mockedFs = fs as jest.Mocked<typeof fs>
+const mockedYaml = yaml as jest.Mocked<typeof yaml>
+
+mockedFs.readFileSync = jest
+  .fn()
+  .mockReturnValue('parsed config in json or yaml format')
+mockedYaml.load = jest.fn().mockReturnValue({ parsed: 'yaml' })
+const pathToFile = 'some/path/to/file.yaml'
+
+import {
+  readFileToString,
+  isJson,
+  getFileFromPath,
+  parseToJson,
+} from './get-file-from-path'
+
+describe('readFileToString()', () => {
+  it('reads correct path', () => {
+    readFileToString(pathToFile)
+
+    expect(mockedFs.readFileSync).toHaveBeenCalledWith(pathToFile)
+  })
+
+  it('always returns string', () => {
+    mockedFs.readFileSync.mockReturnValueOnce(Buffer.from('some buffer'))
+
+    expect(readFileToString(pathToFile)).toBe('some buffer')
+  })
+})
+
+describe('isJson()', () => {
+  it('returns true when passed filePath ends with .json', () => {
+    expect(isJson('/dev/path/to/some/json/file.json')).toBe(true)
+  })
+
+  it('returns false when passed filePath does not end with .json', () => {
+    expect(isJson('/dev/path/to/some/json/file.yml')).toBe(false)
+  })
+})
+
+describe('parseToJson()', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('parses YAML if YAML file is passed', () => {
+    parseToJson('/path/to/json/file.yml')('file: "yaml"')
+
+    expect(mockedYaml.load).toHaveBeenCalledTimes(1)
+  })
+
+  it('parses not YAML if JSON file is passed', () => {
+    parseToJson('/path/to/json/file.json')('{}')
+
+    expect(mockedYaml.load).toHaveBeenCalledTimes(0)
+  })
+})
+
+describe('getFileFromPath()', () => {
+  const innerParse = jest.fn().mockReturnValue({ parsed: 'tojson' })
+
+  beforeAll(() => {
+    readFileToString = jest.fn().mockReturnValue('string')
+    parseToJson = jest.fn(() => innerParse)
+  })
+
+  it('reads the file to string', () => {
+    getFileFromPath('some/path/config.yaml')
+
+    expect(readFileToString).toHaveBeenCalledWith('some/path/config.yaml')
+  })
+
+  it('parses the string to json', () => {
+    getFileFromPath('some/path/config.yaml')
+
+    expect(parseToJson).toHaveBeenCalledWith('some/path/config.yaml')
+    expect(innerParse).toHaveBeenCalledWith('string')
+  })
+
+  it('returns the expected file', () => {
+    const result = getFileFromPath('some/path/config.yaml')
+
+    expect(result).toStrictEqual({
+      filePath: 'some/path/config.yaml',
+      contents: { parsed: 'tojson' },
+    })
+  })
+})
--- a/src/utils/get-file-from-path.ts
+++ b/src/utils/get-file-from-path.ts
@@ -1,0 +1,26 @@
+import R from 'ramda'
+import fs from 'fs'
+import yaml from 'js-yaml'
+
+import { File } from './read-file'
+
+export const readFileToString = (filePath: string): string =>
+  fs.readFileSync(filePath).toString()
+
+export const isJson = R.compose<string, string[], string, boolean>(
+  R.equals('json'),
+  R.last,
+  R.split('.')
+)
+
+export const parseToJson = (filePath: string) => (
+  fileAsString: string
+): JSONObject =>
+  isJson(filePath) ? JSON.parse(fileAsString) : yaml.load(fileAsString)
+
+export const getFileFromPath = (filePath: string): File =>
+  R.compose<string, string, JSONObject, File>(
+    contents => ({ contents, filePath }),
+    parseToJson(filePath),
+    readFileToString
+  )(filePath)
--- a/src/utils/read-file.test.ts
+++ b/src/utils/read-file.test.ts
@@ -1,43 +1,52 @@
-jest.mock('glob')
-jest.mock('fs')
-jest.mock('js-yaml')
-import glob from 'glob'
-import fs from 'fs'
-import yaml from 'js-yaml'
-const mockedGlob = glob as jest.Mocked<typeof glob>
-const mockedFs = fs as jest.Mocked<typeof fs>
-const mockedYaml = yaml as jest.Mocked<typeof yaml>
+jest.mock('./find-files')
+jest.mock('./get-file-from-path')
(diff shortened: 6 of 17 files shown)


src/load.test.ts
```diff
@@ -92,7 +92,10 @@ describe('load behaves as expected', () => {
   it('reads the config based on process.env.RUNTIME_ENVIRONMENT', () => {
     load()
 
-    expect(mockedReadConfigFile).toHaveBeenCalledWith(RUNTIME_ENVIRONMENT)
+    expect(mockedReadConfigFile).toHaveBeenCalledWith(
+      expect.any(String),
```
chapati23, line 96: Why the extra `expect` here? Maybe this is some jest magic I'm not familiar with but the normal signature of  `toHaveBeenCalledWith` is `toHaveBeenCalledWith(arg1, arg2, ...)`.

so if you expect `mockedReadConfigFiles` to be called with two args, why not just `toHaveBeenCalledWith(basePath, RUNTIMIE_ENVIRONMENT)`?

src/utils/find-files.test.ts
```diff
@@ -0,0 +1,103 @@
...
+
+const mockedFileNames = ['/dev/config/development.yml', 'schema.json']
+const mockedResolvedPath = '/dev'
+mockedGlob.sync = jest.fn().mockReturnValue(mockedFileNames)
+mockedPath.resolve = jest.fn().mockReturnValue(mockedResolvedPath)
+mockedGetFileExtensionPattern.mockReturnValue('json')
+
+import { findConfigFiles, findFiles, isNotSchema } from './find-files'
+
+describe('findFiles works as expected', () => {
```
chapati23: nit: "works as expected" sounds very redunant IMO. every test essentially is about checking whether code works as expected.

IMO we could just use the function names for these util `describe` statements.

```suggestion
describe('findFiles()', () => {
```

src/utils/find-files.ts
```diff
@@ -0,0 +1,31 @@
+import R from 'ramda'
+import path from 'path'
+import glob from 'glob'
+
+import { getFileExtensionPattern } from './read-file'
+
+export const isNotSchema = (filePath: string): boolean =>
+  R.compose<string, string[], string, boolean>(
+    fileName => !fileName.startsWith('schema'),
```
chapati23: think positive :)

`isSchema` makes the code a bit easier IMO, because we already have a way of saying `not` with code in form of `!`:

```suggestion
    fileName => fileName.startsWith('schema'),
```

src/utils/find-files.ts
```diff
@@ -0,0 +1,31 @@
+import R from 'ramda'
+import path from 'path'
+import glob from 'glob'
+
+import { getFileExtensionPattern } from './read-file'
+
+export const isNotSchema = (filePath: string): boolean =>
```
chapati23:
```suggestion
export const isSchema = (filePath: string): boolean =>
```

src/utils/find-files.ts
chapati23: Also: Do we want to leave the option to give different names to the schema? Currently `schemablablabla.json` would pass here.

```suggestion
    fileName => fileName === 'schema.json',
```

would be the safest option

src/utils/find-files.ts
```diff
@@ -0,0 +1,31 @@
...
+
+  return glob.sync(globPattern, { cwd: path.resolve(basePath), absolute: true })
+}
+
+export const findConfigFiles = (
+  basePath: string,
+  fileName?: string
+): string[] =>
+  R.filter(
+    isNotSchema,
```
chapati23:
```suggestion
    !isSchema,
```

src/utils/find-files.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('invokes glob.sync', () => {
```
chapati23, line 26: this test feels superfluous given that you also do `it('invokes glob.sync with correct default glob')`

src/utils/find-files.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    findFiles('some/path')
+
+    expect(mockedGlob.sync).toHaveBeenCalledTimes(1)
+  })
+
+  it('returns result of glob.sync', () => {
```
chapati23, line 32: generally you should aim to make your unit tests less implementation specific and more results-oriented. what we're interested in is mainly that our code delivers the correct result, not all the invocations that it does to deliver it.

IMO a lot of the below tests could be replaced with two simpler tests:

```js
  it('returns all config files in a given directory')
```

```js
  it('filters out the schema file')
```

src/utils/find-files.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    findFiles('some/path')
+
+    expect(mockedGlob.sync).toHaveBeenCalledWith(expect.any(String), {
+      cwd: mockedResolvedPath,
+      absolute: true,
+    })
+  })
+})
+
+describe('isNotSchema works as expected', () => {
```
chapati23:
```suggestion
describe('isSchema()', () => {
```

src/utils/find-files.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  it('returns true for paths where the filename does not start with schema', () => {
```
chapati23:
```suggestion
  it('returns false for paths where the filename does not start with schema', () => {
```

src/utils/find-files.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    expect(isNotSchema('/dev/config/development.yaml')).toBe(true)
```
chapati23:
```suggestion
    expect(isSchema('/dev/config/development.yaml')).toBe(false)
```

src/utils/find-files.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  })
+
+  it('returns false for paths where the filename starts withs schema', () => {
```
chapati23:
```suggestion
  it('returns true for paths where the filename starts withs schema', () => {
```

src/utils/find-files.test.ts
```diff
@@ -0,0 +1,103 @@
...
+    expect(isNotSchema('/dev/config/schema.json')).toBe(false)
```
chapati23:
```suggestion
    expect(isSchema('/dev/config/schema.json')).toBe(true)
```

src/utils/find-files.test.ts
```diff
@@ -0,0 +1,103 @@
...
+  })
+})
+
+describe('findConfigFiles works as expected', () => {
```
chapati23:
```suggestion
describe('findConfigFiles()', () => {
```

src/utils/get-file-from-path.test.ts
```diff
@@ -0,0 +1,94 @@
...
+const pathToFile = 'some/path/to/file.yaml'
+
+import {
+  readFileToString,
+  isJson,
+  getFileFromPath,
+  parseToJson,
+} from './get-file-from-path'
+
+describe('readFileToString works as expected', () => {
```
chapati23:
```suggestion
describe('readFileToString()', () => {
```

src/utils/get-file-from-path.ts
```diff
@@ -0,0 +1,26 @@
+import R from 'ramda'
+import fs from 'fs'
+import yaml from 'js-yaml'
+
+import { File } from './read-file'
+
+export const readFileToString = (filePath: string): string =>
```
chapati23, line 7: This is premature optimisation IMO. You're using this util function in _one_ place and it's just a wrapper for a simple built-in node.js function.

I'd just get rid off it, that way you also save some LOC in your tests

src/utils/get-file-from-path.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  })
+
+  it('always returns string', () => {
+    mockedFs.readFileSync.mockReturnValueOnce(Buffer.from('some buffer'))
+
+    expect(readFileToString(pathToFile)).toBe('some buffer')
+  })
+})
+
+describe('isJson works as expected', () => {
```
chapati23:
```suggestion
describe('isJson()', () => {
```

src/utils/get-file-from-path.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  it('returns true when passed filePath ends with .json', () => {
+    expect(isJson('/dev/path/to/some/json/file.json')).toBe(true)
+  })
+
+  it('returns false when passed filePath does not end with .json', () => {
+    expect(isJson('/dev/path/to/some/json/file.yml')).toBe(false)
+  })
+})
+
+describe('parseToJson works as expected', () => {
```
chapati23:
```suggestion
describe('parseToJson()', () => {
```

src/utils/get-file-from-path.test.ts
```diff
@@ -0,0 +1,94 @@
...
+  })
+
+  it('parses not YAML if JSON file is passed', () => {
+    parseToJson('/path/to/json/file.json')('{}')
+
+    expect(mockedYaml.load).toHaveBeenCalledTimes(0)
+  })
+})
+
+describe('getFileFromPath works as expected', () => {
```
chapati23:
```suggestion
describe('getFileFromPath()', () => {
```

src/utils/read-file.test.ts
```diff
@@ -1,75 +1,66 @@
...
+  contents: {
+    parsed: 'values',
+  },
+  filePath: mockedConfigFilePaths[0],
+}
+mockedGetFileFromPath.mockReturnValue(mockedFile)
+
+import { readConfigFile, readSchemaFile } from './read-file'
 
 describe('readConfigFile behaves as expected', () => {
```
chapati23:
```suggestion
describe('readConfigFile()', () => {
```

src/utils/read-file.ts
```diff
@@ -1,95 +1,59 @@
...
-      `None of ${filename}.${getFileExtensionPattern()} found. One of these files must exist.`
+  if (R.isEmpty(filePaths)) {
+    throw new Error(
+      `None of ${fileName}.${getFileExtensionPattern()} found. One of these files must exist.`
     )
-  } else if (files.length > 1) {
+  } else if (filePaths.length > 1) {
     throw new Error(
-      `More than one of ${filename}.${getFileExtensionPattern()} found. Exactly one must exist.`
+      `More than one of ${fileName}.${getFileExtensionPattern()} found. Exactly one must exist.`
```
chapati23, line 29: not sure i'm reading this correctly. but there can never be two `dev.yaml` files in the same directory in most common file systems

src/validate.ts
```diff
@@ -1,9 +1,50 @@
...
+import { findConfigFiles } from './utils/find-files'
+import { getFileFromPath } from './utils/get-file-from-path'
+import { validateConfig } from './utils/validate-config'
 
-export const validate = (schema: Schema): true => {
-  if (R.isNil(schema)) {
-    throw new Error('Config does not match the specified schema')
+const validateConfigAgainstSchema = (schema: Schema) => (
+  configFilePath: string
+): true => {
```
chapati23, line 10: maybe a stupid question, but shouldn't this be `boolean` instead of `true`?

src/validate.ts
```diff
@@ -1,9 +1,50 @@
...
+  const configFile = getFileFromPath(configFilePath)
+
+  return validateConfig(configFile.contents, schema)
+}
+
+export const validate = (
+  schemaPath: string,
+  ...configPaths: string[]
+): true => {
```
chapati23, line 19: same question here

chapati23 (changes requested):
Great work.

Some code comments. And two general feedback items:

**1. MR Size**
Try to keep your MRs smaller. The MR description plus title imply that it's mostly about `validate()` when in reality the large majority of the code changes were refactorings that didn't have much to do with this.

It took me almost an hour to complete the review (incl. running code locally) which is IMO at the border of "too long".

**2. Unit Tests Approach**
First of all: 👍 for writing lots of tests, appreciate it.
What I think you can improve upon is the types of tests you write. I saw a lot of very low-level tests that were very implementation focussed. This is often a sign of "make the tests fit the solution" and results in lots of LOC to maintain which don't always provide high value.

Try to switch your mindset a bit more towards TDD, where you don't have the solution yet. This forces you to write your tests more focussed on the result that you want from your code, and not the specific implementation of it.

src/load.test.ts
mohoff: it's two different kind of `expect`s. The second one is a helper that is used a placeholder to be happy with 'any kind of string'.

The idea is to `expect` on things that a particular test is validating. Here, it's checking if the `RUNTIME_ENVIRONMENT` is passed correctly. Thus, `mockedReadConfigFiles` must be called with (<any string we don't care about in this test>, RUNTIME_ENVIRONMENT)`

src/utils/find-files.test.ts
mohoff: fine with me

src/utils/find-files.ts
mohoff: Regarding your fist comment: Seems a nit to me since we only check for the negated form. Now, the `!` is in `!isSchema`. However, I think i'm gonna replace `R.filter` with `R.reject` :)

Regarding your second comment: Yeah but i'd say we leave it for now as this will change once we can pass a configuration object. One of these option will be the schema path. So this will change anyway.

src/utils/read-file.ts
mohoff: but there can be a `dev.yaml` and `dev.yml`, `dev.json` in the same dir, which this check would correctly prohibit. Otherwise, it would mean there are multiple config files for one environment, which isn't supported.

src/validate.ts
mohoff: values can also be types, e.g. you can construct `type Levels = 'error | 'info' | 'debug'` as well. Since this function either throws or returns `true`, the stricter type here is the literal `true` rather than `boolean`.

src/validate.ts
mohoff: answered above

src/utils/find-files.ts
mohoff: Thanks for all the commit suggestions. When I hit this button now, it'll create a new commit with that small change? Guess we never `commit suggestion` but use this feature to make clear what changes we mean.

src/utils/get-file-from-path.ts
mohoff: Hm I don't agree:

- it's function composition.
- it hides the detail of `readFile` eventually returning a Buffer.
- it's explicit about the result. It's a string, not a json/buffer/yaml/`File`-type/etc.
- it's unary, so I can directly use it as part of `R.compose`. `fs.readFileSync` is not unary.
- it's easy to mock. I agree, the tests for this function is a bit dumb but just there for completeness.

src/utils/find-files.test.ts
mohoff: Was thinking that `glob.sync` can be invoked but wrong arguments are passed. In such cases, you can immediately see from the tests that 1. it's called and 2. but with wrong arguments.

Without that test, you would first have to debug if it actually reaches `glob.sync` at all.

Bit nitty discussion but I still think this test _can_ be there.

src/utils/find-files.test.ts
mohoff: making this comment once is sufficient ;) I bring this degree of proactivity :p

src/utils/find-files.test.ts
mohoff: Hm yeah will try to pay some attention.

But we reached a good level of nit now. For me, it's 99% about 1. having tests, 2. testing the right thing, 3. easy to maintain/update

src/utils/find-files.test.ts
mohoff: updated

src/utils/find-files.test.ts
mohoff: updated

src/utils/find-files.test.ts
mohoff: updated

src/utils/find-files.test.ts
mohoff: updated

src/utils/find-files.test.ts
mohoff: updated

src/utils/find-files.test.ts
mohoff: updated

src/utils/find-files.ts
mohoff: updated

src/utils/find-files.ts
mohoff: regarding first comment: updated

src/utils/find-files.ts
mohoff: updated

src/utils/get-file-from-path.test.ts
mohoff: updated

src/utils/get-file-from-path.test.ts
mohoff: updated

src/utils/get-file-from-path.test.ts
mohoff: updated

src/utils/get-file-from-path.test.ts
mohoff: updated

src/utils/read-file.test.ts
mohoff: updated

src/load.test.ts
chapati23: got it, ok

src/utils/find-files.ts
chapati23: maybe i got too excited about this feature. i actually really like this about github's review flow that you can make concrete code suggestions to save some feedback loops on smaller stuff.

however, i expected that it's possible to batch these together and _not_ create a new commit for ever single suggestion:

image

(thought that's what "Add suggestion to batch" was for)

src/validate.ts
chapati23: TIL, thx

src/utils/read-file.ts
chapati23: got it, thx

src/utils/find-files.test.ts
chapati23: Hmm, but doesn't TypeScript protect us against calling this with wrong arguments by requiring them to be string?

src/utils/find-files.test.ts
chapati23: Agree with all of those. My point is, if you couple your tests super tightly to the implementation they are not all testing the right thing (which is mostly that the code delivers the correct result) and become harder to maintain because changing the implementation becomes more costly because you have to update more tests.

Will resolve this conversation now because refactoring all tests would be a bit much. For the next tests, tho, I'd like to challenge you to try out a little bit of TDD and writing some of your tests first, so you experience the difference.

src/utils/get-file-from-path.ts
chapati23: My main counter-argument is that this one-time-only util function costs 20 LOC in tests. To your points:

- it's the most minimal example of function composition i can think off 😛
- it's a simple `.toString()`, IMO this is very clear that it's not a buffer
- `.toString()` is very clear about what it returns already
- ok, fair point
- it's not that much harder to mock `fs.readFileSync` IMO, but ok, small point

I realise this is a nit contest, so I'll resolve this. My main point is: Sometimes the cost of creating abstractions outweighs its benefits e.g. by generating a lot more LOC to maintain.

src/utils/read-file.ts
chapati23: got it 👍

src/utils/find-files.ts
mohoff: I saw that button too but i felt with a full-text search i'm quicker to update all this `describe('... behaves as expected')` spots.

I'll try it next time!

src/utils/find-files.test.ts
mohoff: Not entirely sure. I think it's typesafe at transpile time but not at runtime.

mohoff: After some research, I found that Typescript only offers static analysis, so transpile-time type safety, by itself. With type guards and libraries such as `io-ts`, you can achieve basic runtime type-safety where needed.

I think it's important that we always validate user input, i.e. argument passed to `load()` and `validate()`, and data coming in programmatically (result of sops binary, results of KMS calls, etc.). Maybe worth another GitHub-issue to make a sanity check once we have all basic features in master.

chapati23 (approved):

PR #20, opened 2019-10-09
Refactor/class instantiation

### LINKED ISSUE #7: Make top-level export a Class `StrongConfig`
Currently, we export an object with `{ load, validate }` properties. Instead, we want to export a class named `StrongConfig` that accepts as configuration obj as a constructor argument. Members would expose existing `load()` and `validate()` functionality.

mohoff:
Donezo:

- [x] Export class `StrongConfig` instead of two top-level functions `load()` and `validate()`.
- [x] Move state (loaded config) into `StrongConfig` class instance.
- [x] Move types to `src/types.ts`. I learnt that `.d.ts` files define ambient types for non-typescript code bases. So this was the wrong approach. With the new file, we have to export/import types.
- [x] Update tests accordingly

The refactor to class `StrongConfig` is a BREAKING CHANGE

Fixes https://github.com/strong-config/node/issues/7

### DIFF
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -1,11 +1,57 @@
-import * as strongConfig from './index'
+jest.mock('./load')
+jest.mock('./validate')
 
-describe('strong-config exports are as expected', () => {
-  it('exports a load function', () => {
-    expect(strongConfig.load).toBeInstanceOf(Function)
+import { load } from './load'
+import { validate } from './validate'
+
+const mockedLoad = load as jest.MockedFunction<typeof load>
+const mockedValidate = validate as jest.MockedFunction<typeof validate>
+
+const mockedConfig = { some: 'config', runtimeEnvironment: 'development' }
+const mockedValidationResult = true
+mockedLoad.mockReturnValue(mockedConfig)
+mockedValidate.mockReturnValue(mockedValidationResult)
+
+import StrongConfig from './index'
+
+describe('StrongConfig class', () => {
+  it('StrongConfig is a class that can be instantiated', () => {
+    expect(new StrongConfig()).toBeDefined()
   })
 
-  it('exports a validate function', () => {
-    expect(strongConfig.validate).toBeInstanceOf(Function)
+  it('StrongConfig accepts and stores a configuration object', () => {
+    const mockConfiguration = {}
+    const strongConfig = new StrongConfig(mockConfiguration)
+
+    expect(strongConfig.configuration).toStrictEqual(mockConfiguration)
+  })
+
+  it('strongConfig exposes "load()"', () => {
+    expect(new StrongConfig().load).toBeInstanceOf(Function)
+  })
+
+  it('strongConfig exposes "validate()"', () => {
+    expect(new StrongConfig().validate).toBeInstanceOf(Function)
+  })
+
+  it('strongConfig.load() calls imported load() and returns its result', () => {
+    const strongConfig = new StrongConfig()
+
+    const result = strongConfig.load()
+
+    expect(mockedLoad).toHaveBeenCalledTimes(1)
+    expect(result).toStrictEqual(mockedConfig)
+  })
+
+  it('strongConfig.validate() calls imported validate() and returns its result', () => {
+    const strongConfig = new StrongConfig()
+
+    const result = strongConfig.validate(
+      'path/to/schema.json',
+      'some/config/path'
+    )
+
+    expect(mockedValidate).toHaveBeenCalledTimes(1)
+    expect(result).toStrictEqual(mockedValidationResult)
   })
 })
--- a/src/index.test.ts
+++ b/src/index.test.ts
@@ -15,43 +15,62 @@ mockedValidate.mockReturnValue(mockedValidationResult)
 import StrongConfig from './index'
 
 describe('StrongConfig class', () => {
-  it('StrongConfig is a class that can be instantiated', () => {
+  beforeEach(() => {
+    jest.clearAllMocks()
+  })
+
+  it('can be instantiated', () => {
     expect(new StrongConfig()).toBeDefined()
   })
 
-  it('StrongConfig accepts and stores a configuration object', () => {
-    const mockConfiguration = {}
-    const strongConfig = new StrongConfig(mockConfiguration)
+  it('accepts and stores a parameter object', () => {
+    const mockedParams = {}
+    const strongConfig = new StrongConfig(mockedParams)
 
-    expect(strongConfig.configuration).toStrictEqual(mockConfiguration)
+    expect(strongConfig.params).toStrictEqual(mockedParams)
   })
 
-  it('strongConfig exposes "load()"', () => {
+  it('strongConfig instance exposes "load()"', () => {
     expect(new StrongConfig().load).toBeInstanceOf(Function)
   })
 
-  it('strongConfig exposes "validate()"', () => {
+  it('strongConfig instance exposes "validate()"', () => {
     expect(new StrongConfig().validate).toBeInstanceOf(Function)
   })
 
-  it('strongConfig.load() calls imported load() and returns its result', () => {
-    const strongConfig = new StrongConfig()
+  describe('strongConfig.load()', () => {
+    it('calls imported load() and returns its result', () => {
+      const strongConfig = new StrongConfig()
 
-    const result = strongConfig.load()
+      const result = strongConfig.load()
 
-    expect(mockedLoad).toHaveBeenCalledTimes(1)
-    expect(result).toStrictEqual(mockedConfig)
+      expect(mockedLoad).toHaveBeenCalledTimes(1)
+      expect(result).toStrictEqual(mockedConfig)
+    })
+
+    it('memoizes previously loaded config', () => {
+      const strongConfig = new StrongConfig()
+
+      const firstLoadResult = strongConfig.load()
+      const secondLoadResult = strongConfig.load()
+
+      expect(mockedLoad).toHaveBeenCalledTimes(1)
+      expect(firstLoadResult).toStrictEqual(mockedConfig)
+      expect(secondLoadResult).toStrictEqual(mockedConfig)
+    })
   })
 
-  it('strongConfig.validate() calls imported validate() and returns its result', () => {
-    const strongConfig = new StrongConfig()
+  describe('strongConfig.validate()', () => {
+    it('calls imported validate() and returns its result', () => {
+      const strongConfig = new StrongConfig()
 
-    const result = strongConfig.validate(
-      'path/to/schema.json',
-      'some/config/path'
-    )
+      const result = strongConfig.validate(
+        'path/to/schema.json',
+        'some/config/path'
+      )
 
-    expect(mockedValidate).toHaveBeenCalledTimes(1)
-    expect(result).toStrictEqual(mockedValidationResult)
+      expect(mockedValidate).toHaveBeenCalledTimes(1)
+      expect(result).toStrictEqual(mockedValidationResult)
+    })
   })
 })
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,2 +1,26 @@
-export { load } from './load'
-export { validate } from './validate'
+import { load } from './load'
+import { validate } from './validate'
+
+interface Configuration {
+  // TODO: define Configuration interface
+}
+
+export default class StrongConfig {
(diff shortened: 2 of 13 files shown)


chapati23:
Thanks, have you tried this out in an example application?

Most importantly: Can you confirm that we're executing the relatively expensive `fs.readFileSync` operation only once even when multiple files of a consuming application make use of the `config` object?

mohoff:
Looks like I can't make a `Draft PR` out of this anymore...

Yes, tested for both commonjs and es6 import style.

I realized that the config state should be placed in the class now instead of the load-module. Will update this in this PR

mohoff:
Just pushed and updated the PR description. Ready for review.

src/index.test.ts
```diff
@@ -1,11 +1,72 @@
...
+    jest.clearAllMocks()
+  })
+
+  it('StrongConfig is a class that can be instantiated', () => {
+    expect(new StrongConfig()).toBeDefined()
   })
 
-  it('exports a validate function', () => {
-    expect(strongConfig.validate).toBeInstanceOf(Function)
+  it('StrongConfig accepts and stores a parameter object', () => {
```
chapati23: nit: no need to repeat `strongconfig` in every single `it()` statement. you already wrote it in the `describe()` blog.
IMO it-statements should always read like a full english sentence, e.g.:

```js
describe('StrongConfig class', () => { // this sets the context that all following it-statements refer to
  it('accepts and stores a parameter object', () => {})
})
```

src/index.test.ts
```diff
@@ -1,11 +1,72 @@
...
+  })
+
+  it('strongConfig exposes "load()"', () => {
+    expect(new StrongConfig().load).toBeInstanceOf(Function)
+  })
+
+  it('strongConfig exposes "validate()"', () => {
+    expect(new StrongConfig().validate).toBeInstanceOf(Function)
+  })
+
```
chapati23, line 40: here you're beginning a new context, so you should wrap the below in a new `describe()`

```suggestion
describe('load()', () => {
```

src/index.test.ts
```diff
@@ -1,11 +1,72 @@
...
+  it('strongConfig.load() calls imported load() and returns its result', () => {
```
chapati23: and then again you can skip repeating the context in it-statements every time

```suggestion
  it('calls imported load() and returns its result', () => {
```

src/index.test.ts
```diff
@@ -1,11 +1,72 @@
...
+    const strongConfig = new StrongConfig()
+
+    const result = strongConfig.load()
+
+    expect(mockedLoad).toHaveBeenCalledTimes(1)
+    expect(result).toStrictEqual(mockedConfig)
+  })
+
+  it('strongConfig.load() caches previously loaded config', () => {
```
chapati23: This for example is a great test IMO. It clearly tests an actual _feature_ of the code (not implementation details like which function called what other util function)

src/index.ts
```diff
@@ -1,2 +1,35 @@
...
+import { load } from './load'
+import { validate } from './validate'
+
+import { MemoizedConfig } from './types'
+
+interface Parameters {
+  // TODO: define Parameters interface
+}
+
+export = class StrongConfig {
```
chapati23, line 10: ts noob question: why the `=`? in pure JS you would just do `export class StrongConfig`, is there different behavior in TS?

src/index.test.ts
chapati23: nit: the config object is named `MemoizedConfig`, so it'd be more consistent to always use `memoize` (instead of `cache`)

tsconfig.json
```diff
@@ -13,12 +13,13 @@
     "noImplicitAny": true,
     "outDir": "./lib",
     "paths": {
-      "*": ["node_modules/*"]
+      "*": ["node_modules/*", "src/*"]
```
chapati23: what does this do?

tsconfig.json
```diff
@@ -13,12 +13,13 @@
...
     },
     "resolveJsonModule": true,
     "sourceMap": true,
     "strict": true,
-    "target": "es6"
+    "target": "es6",
+    "lib": ["es6"]
```
chapati23, line 22: what does this do?

src/utils/read-file.ts
```diff
@@ -4,15 +4,7 @@ import path from 'path'
...
-export enum FileExtension {
-  JSON = 'json',
-  YAML = 'yaml',
-  YML = 'yml',
-}
-export type File = {
-  contents: EncryptedConfig | Schema
-  filePath: string
-}
+import { File, FileExtension } from '../types'
```
chapati23: are these types used anywhere else or just here?

always a bit wary of throwing too much into these `types.{js|ts}` kind of files as they can grow into black holes over time in my experience

chapati23 (changes requested):
Good work. Left a few comments.

src/index.ts
mohoff: It works with both commonjs and es6-style imports. Without this special `=` syntax, you'd had to do `const StrongConfig = require('strong-config/node').default`. 

Docs: https://www.typescriptlang.org/docs/handbook/modules.html#export--and-import--require

tsconfig.json
mohoff: wanted to make imports a la `NODE_PATH` work. `src` is included as path for module resolution. It worked but messed up jest badly, so I went back to relative imports. Good point, we can remove it again.

tsconfig.json
mohoff: Basically, the build output in `lib/` should be smaller with this as it generates code for more than one target by default.

For the future, can I kindly ask you to read the docs for these kind of questions? https://www.typescriptlang.org/docs/handbook/compiler-options.html

src/utils/read-file.ts
mohoff: Files is used twice, FileExtension once.

I'm also not sure about this. Before going the `types.ts` way, I did some research to check how other TS projects are doing this. But it seemed to me that everyone is doing freestyle (or I don't understand their indivdual requirements well enough).

When `types.ts` becomes too cramped, we can do

```
types/
    io.ts
    config.ts
    ...
```

but same here -- lacking experience to make an educated decision for now.

src/index.test.ts
mohoff: updated

src/index.test.ts
mohoff: updated

src/index.test.ts
mohoff: updated

src/index.test.ts
mohoff: updated

tsconfig.json
mohoff: removed

src/index.test.ts
chapati23: don't see the update :)

mohoff: sorry, pushed now

src/index.ts
chapati23: thanks for the context. looking at the link you sent, what's the reason we can't use the same import/export  mechanism as we do everywhere else?

```ts
export default class StrongConfig {}
```

```ts
import StrongConfig from '.'
```

relevant part of the docs:

image

mohoff: Because of the reason I gave: The commonjs would then look like this:

```js
const StrongConfig = require('strong-config/node').default
```

That's not nice. And `export =` fixes that so it's the syntax you gave for es6-imports and for commonjs imports it is:

```js
const StrongConfig = require('strong-config/node')
```

tsconfig.json
chapati23: Look, I'm perfectly capable to read docs and learn the entirety of TypeScript syntax and best practices. My limitation is time. If I don't understand why a certain change is in as PR and how it supports the core PR goal, then I will just ask. I could have asked a little more clearly, though:

> Why is this change in this PR?

IMO the PR creator has the responsibility to make the review as easy as possible. The harder you make it for reviewers, and the more digging you require them to do, the slower your turnaround time for PRs will be. My main reason for challenging you on these smaller things is that you tend to just throw them into unrelated PRs without any comment or explanation.

This way you're putting more burden on the reviewer to think about
- Why is this change is in this PR?
- Is it required for the core goal of the PR?
- If not, why is it in this PR then?

Especially given that this is our first TypeScript project, and you're the first one that gets to play around with it in detail, IMO you also have a knowledge-sharing and knowledge-building responsibility to proactively disseminate your newly acquired knowledge amongst the team.

chapati23: Don't get me wrong, I'm not saying that you can't throw in a one-liner or two of refactoring into otherwise unrelated PRs. It would be a different story for me if there was a line in the PR description that says:

> ### Refactoring
> Added [`lib`](https://www.typescriptlang.org/docs/handbook/compiler-options.html) parameter to typescript config to improve...

mohoff: Ok, I get your point. Will be more careful with these tiny changes in the future.

From my side:
- It's the 4th PR of this project. I'd appreciate if we can work with the granularity of issues for now and let the ts-setup stuff mature over time. As you know, it's also first time typescript for me, so there's a lot of playing around involved, see `.d.ts` topic. I don't think it's a problem to regard questions 'why is this ts-config param set to true and not false?' not important for now.
- I'm happy to share my knowledge at any time.
- I'd appreciate a bit more autonomy and trust in this project.

src/index.ts
chapati23: apologies, you're right. didn't get this at first. have used es6 imports/exports for too long now that i forget the old world still exists :)

src/utils/read-file.ts
chapati23: what's the reason we can't handle it the same way as we're doing with flow currently where we define the types in the same file as the actual code? e.g. in flow we would do:

```js
// read-file.js

type FileExtension = 'json' | 'yml'

function something(extension: FileExtension) {
  somelogic()
}
```

mohoff: Works too! However, I use the type to build up an array of file extensions. To do this, I found that's it's much easier/more convenient to enumerate an enum instead of a literal union type:

- literal union type to array: https://stackoverflow.com/questions/44480644/typescript-string-union-to-string-array

- enum to array: `Object.keys(FileExtension)`. Or with named enums: `Object.values(FileExtension`. See `src/read-file.ts` L10.

tsconfig.json
chapati23: If you feel mistrusted then I'm sorry. That is definitely not the case. My two main drivers for being so involved in strong-config MRs are:

1) Take part in learning about TypeScript
I unfortunately don't have the time to write a lot of code or read all the TypeScript docs anymore. So for me learning these days often happens in MRs. When I see new ways of doing things. If there was small explanation or little comment then this would go a long way towards knowledge sharing for me :)

2) Setting a high quality bar for PRs.
This is our first real public project, so it heavily contributes to defining our reputation as an engineering team. When I see a PR or issue, I'm always trying to put myself into the shoes of three personas:

- Potential users of the package
   Goes without saying, this is the normal code review process to see whether changes make sense for future users of this package. And I think you're mostly doing a great job here which is why most of my comments have been focussed more on the other two personas.
- Engineers that evaluate joining Brickblock
   Top talent is hard to get. They think twice before joining any company and do their due diligence. If they see messy code or communication in publicly available code or PRs or Issues of a company, that decreases our chances to attract them (obviously none of the code here is messy, but you get the idea)
- External contributors
   If someone creates their first PR for an open source project, they usually look at past PRs as an example for standards around PR description, code quality, communication, etc.

src/utils/read-file.ts
chapati23: Ok, then i'd suggest the following:

- For widely shared general types, we gather them in some common place (e.g. `types.js`.)
- For types that are very tightly coupled to a specific file, we define them inside that file and import/export them from there

What we should avoid IMO is to re-create our file structure under `src/` in `types/`, e.g. avoid this:

```
- src
   - config.ts
   - read-file.ts
   - some-helper.ts
   - validate.ts
- types
   - config.ts
   - read-file.ts
   - some-helper.ts
   - validate.ts
```

mohoff: Alright. I put `File` and `FileExtension` back into `read-file.ts`. The other types i'd leave in `types.ts` for now as these are rather global + often imported ones.

chapati23 (approved):
🙌

PR #21, opened 2019-10-10
Chore/impove gitlab ci

### LINKED ISSUE #13: Improve Gitlab CI pipeline
Once we are making use of the new `extend` keyword in the `.gitlab-ci.yml` in our pipeline files, let's adopt the pipeline setup of this project accordingly.

The goal is to have
- efficient pipelines than run fast and make use of caching strategies via optimal configurations for `cache` and `artificats`
- Simplify and DRY out the yaml

mohoff:
Donezo:

- [x] Include more jobs that replace custom ones.

Only the `release` job remains as custom definition. This is because contained variables are specific to Github.

Need to merge #20 first as I branched off it.

fixes #13

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -6,58 +6,20 @@ stages:
   - build
   - release
 
-# Fallback to support `includes`
-cache:
-  key: $CI_COMMIT_REF_SLUG
-  paths:
-    - node_modules/
-    - lib/
-
-.shared:
-  artifacts:
-    paths:
-      - lib/
-      - package.json
-      - CHANGELOG.md
-    expire_in: 30 days
-  cache:
-    key: $CI_COMMIT_REF_SLUG
-    paths:
-      - node_modules/
-      - lib/
-
-build_npm_package:
-  stage: build
-  extends: .shared
-  except:
-    refs:
-      - schedules
-      - tags
-    variables:
-      - $CI_COMMIT_MESSAGE =~ /\[ci-release\]/
-  script:
-    - yarn build
-  cache:
-    policy: pull
-
+# We can't include the release job (https://git.brickblock.sh/devops/project-templates/raw/master/gitlab-ci/release.yml)
+# because it is tailored to our GitLab-hosted infrastructure. This overwrite solves the release with a GitHub project.
 release:
   stage: release
-  extends: .shared
-  cache:
-    policy: pull
-  artifacts: {}
+  extends:
+    - .except-ci-release
+    - .cache-pull-only
   only:
     - master
-  except:
-    refs:
-      - schedules
-      - tags
-    variables:
-      - $CI_COMMIT_MESSAGE =~ /\[ci-release\]/
+  dependencies:
+    - build_npm_package
   script:
     - git config --global user.name "${GIT_USER_NAME}"
     - git config --global user.email "${GIT_USER_EMAIL}"
-    - git config --global push.followTags true
     - git config --global push.default current
     - yarn release
     - echo '//registry.npmjs.org/:_authToken=${SECRET_NPM_TOKEN}'>.npmrc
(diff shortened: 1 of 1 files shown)


src/index.test.ts
```diff
@@ -1,11 +1,72 @@
-import * as strongConfig from './index'
```
chapati23: why has this file (and all the below) changed? has nothing to do with PR description or title

.gitlab-ci.yml
```diff
@@ -6,58 +6,21 @@ stages:
...
     - master
-  except:
-    refs:
-      - schedules
-      - tags
-    variables:
-      - $CI_COMMIT_MESSAGE =~ /\[ci-release\]/
+  artifacts:
+    paths:
+      - CHANGELOG.md
```
chapati23: sorry, i keep forgetting. why do we need this artifact here?

.gitlab-ci.yml
```diff
@@ -6,58 +6,21 @@ stages:
...
-      - schedules
-      - tags
-    variables:
-      - $CI_COMMIT_MESSAGE =~ /\[ci-release\]/
-  script:
-    - yarn build
-  cache:
-    policy: pull
-
 release:
```
chapati23, line 11: i'm sure there's a good reason but could you highlight what's keeping us from just using our shared `release.yml` from the project templates?

chapati23 (changes requested):
thanks, two small comments on the CI stuff.

please remove the other files from this PR, tho. they aren't mentioned anywhere in PR description or title and the changes seem unrelated to CI.

src/index.test.ts
mohoff: As mentioned on Slack, we need to merge in !20 first. After the rebase, this PR is ready for review

chapati23: ah ok, better put that in the PR description so reviewers don't need to switch between two communication streams (plus this wouldn't work with external contributors)

.gitlab-ci.yml
mohoff: No, valid question. Since it's the last stage in this pipeline, we can remove it actually 👍

src/index.test.ts
mohoff: I'm well aware. Since you're the only reviewer right now, I thought this works

.gitlab-ci.yml
mohoff: updated

.gitlab-ci.yml
mohoff: added a comment :)

chapati23: i see, is there no way to extend from the common `release` and then just override the `variables` with the github ones?

.gitlab-ci.yml
chapati23: great, i think the same comment then also applies to https://git.brickblock.sh/devops/project-templates/raw/master/gitlab-ci/release.yml (just as a sidenote)

.gitlab-ci.yml
mohoff: In theory yes. The problem though is that we'll end up with two `release` jobs: the included one and the custom one that extends from the former. That would result in a double-release and a race-condition where the 2nd `git push` would fail due to conflicts.

If you have an idea how to solve this, I'm all ears

chapati23: Hmm, what if we made `release` an extend `.release` instead? Then in this `.gitlab-ci.yml` the `include release.yml` wouldn't trigger the job anymore. 

Instead you'd define the `release` job yourself and extend from the include.

Do you understand what I'm getting at?

mohoff: Understood, but it would only work when we move `.release` to `shared.yml`.

If we don't do that and leave it in `release.yml`, all our existing projects would have to extend from that `.release`. Moving the extension to the individual `.gitlab-ci.yml`s is a step in the wrong direction imo. If we extend inside the same file where `.release` resides, we end up with two release jobs.

chapati23: ok, feels like it's getting too complicated to be DRY here. let's just leave it as is for now ✅

src/index.test.ts
chapati23: after having merged the previous PR, it seems github is doing a worse job than gitlab in dealing with PRs that rely on other PRs.

e.g. it doesn't allow you to rebase a PR on top of master from the UI

so i think if possible we should always try to start new features based off `master` (exceptions can happen when features build upon other features in previous PRs, but this isn't the case here)

mohoff: Agree

chapati23 (approved):
👍

PR #22, opened 2019-10-11
chore: setup example configs with scripts for development

### LINKED ISSUE #19: Create 'example' folder to speed up local development
## tl;dr
Create an `example` folder that is its own self-contained app using `@strong-config/node` to allow faster iteration during development.

## The Problem
Currently, when we want to try out changes on real schemas or config files we have three options:

1. Switch to another app that's already using `strong-config` such as our internal `api-gateway`
   - This is bad because it's tightly coupled with our own infra and external contributors don't have that option.
    - It's also inconvenient in terms of DX to have to constantly switch between two repos just to try out some changes
1. Rely solely on tests
   - Obviously we should always write tests and our tests should be robus
   - But sometimes you just want to try things out quickly, e.g. see how a certain code change affects the transformation of `schema.json` into TypeScript and always having to write a test for this is very cumbersome
1. Create a new app that uses `@strong-config/node`
   - This is of course possible but also very repetitive
   - Have to repeat various steps every time (e.g. create folder, `yarn init`, `yarn add`, manually write schema file, config files...)

## The Proposed Solution
Create an `example` folder inside this repo that:
- [ ] Has one (or several) example `schema.json` to quickly try out how our code behaves in various scenarios
- [ ] Has several example config files to quickly try out how our code behaves in various scenarios

In our `package.json`, we will then create the following yarn tasks:
- [ ] `yarn dev:load $env` - Load a config file. Translate it into TypeScript. Log the result to the console.
- [ ] `yarn dev:load:watch $env` - Same as above, but with hot-reloading on every file change via `nodemon`
- [ ] `yarn dev:validate $env` validates a config yaml against a schema
- [ ] `yarn dev:validate:watch $env` - Same as above, but with hot-reloading on every file change via `nodemon`
- [ ] [BONUS] `yarn dev:encrypt $env` - Encrypts a config yaml
- [ ] [BONUS] `yarn dev:decrypt $env` - Decrypts a config yaml

mohoff:
Donezo:

- [x] Added example config and schema under `/example`
- [x] Added various scripts to support development for `load()` and `validate()`, inluding 'watch'-functionality that refreshes on files changes by using `nodemon`.
- [x] Added `time` to these commands so you can see differences on how fast the config loads or how fast is it validated.
- [x] Corrected some tests that were broken: Fixed type import. Fixed a config path param

How to test:
Run any of the new commands and make changes to the example config or schema. The `:watch` commands update on-change and show the newly loaded config, or, if the config still validates against the schema.

fixes #19

### DIFF
--- a/nodemon.json
+++ b/nodemon.json
@@ -1,0 +1,5 @@
+{
+  "ignore": ["**/*.test.ts"],
+  "watch": ["src", "example"],
+  "ext": "ts,json,yaml,yml"
+}
--- a/package.json
+++ b/package.json
@@ -23,10 +23,11 @@
     "watch": "tsc --watch",
     "clean": "rimraf build/",
     "build": "tsc",
-    "dev:load:commonjs": "export RUNTIME_ENVIRONMENT=development; yarn build && echo 'Loaded Config:\n' && time node -p -e 'const StrongConfig = require(\".\"); JSON.stringify((new StrongConfig()).load(\"./example\"))' | jq .",
-    "dev:load:es6": "export RUNTIME_ENVIRONMENT=development; echo 'Loaded Config:\n' && time ts-node -T -p -e 'import StrongConfig from \".\"; JSON.stringify((new StrongConfig()).load(\"example\"))' | jq .",
+    "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time node -p -e 'const StrongConfig = require(\".\"); JSON.stringify((new StrongConfig()).load(\"./example\"))' | jq .",
+    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T -p -e 'import StrongConfig from \".\"; JSON.stringify((new StrongConfig()).load(\"example\"))' | jq .",
+    "dev:load": "yarn dev:load:es6",
     "dev:load:watch": "nodemon --exec 'yarn dev:load:es6'",
-    "dev:validate": "export RUNTIME_ENVIRONMENT=development; echo 'Validation Result:\n' && time ts-node -T -p -e 'import StrongConfig from \".\"; (new StrongConfig()).validate(\"example/schema.json\", \"example\")'",
+    "dev:validate": "echo 'Validation Result:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T -p -e 'import StrongConfig from \".\"; (new StrongConfig()).validate(\"example/schema.json\", \"example\")'",
     "dev:validate:watch": "nodemon --exec 'yarn dev:validate'"
   },
   "devDependencies": {
@@ -44,6 +45,7 @@
     "@typescript-eslint/eslint-plugin": "^2.3.1",
     "@typescript-eslint/parser": "^2.3.1",
     "commitlint": "^8.2.0",
+    "cross-env": "^6.0.3",
     "eslint": "^6.4.0",
     "eslint-config-prettier": "^6.3.0",
     "eslint-plugin-import": "^2.18.2",
--- a/package.json
+++ b/package.json
@@ -22,7 +22,12 @@
     "todo": "leasot '**/*.ts' --ignore 'node_modules/**/*','lib/**/*','.git/**/*' || true",
     "watch": "tsc --watch",
     "clean": "rimraf build/",
-    "build": "tsc"
+    "build": "tsc",
+    "dev:load:commonjs": "export RUNTIME_ENVIRONMENT=development; yarn build && echo 'Loaded Config:\n' && time node -p -e 'const StrongConfig = require(\".\"); JSON.stringify((new StrongConfig()).load(\"./example\"))' | jq .",
+    "dev:load:es6": "export RUNTIME_ENVIRONMENT=development; echo 'Loaded Config:\n' && time ts-node -T -p -e 'import StrongConfig from \".\"; JSON.stringify((new StrongConfig()).load(\"example\"))' | jq .",
+    "dev:load:watch": "nodemon --exec 'yarn dev:load:es6'",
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -22,7 +22,12 @@
     "todo": "leasot '**/*.ts' --ignore 'node_modules/**/*','lib/**/*','.git/**/*' || true",
     "watch": "tsc --watch",
     "clean": "rimraf build/",
-    "build": "tsc"
+    "build": "tsc",
+    "dev:load:commonjs": "export RUNTIME_ENVIRONMENT=development; yarn build && time node -e 'const StrongConfig = require(\".\"); console.log(\"Example Config:\", (new StrongConfig()).load(\"./example\"))'",
```
chapati23: small improvement suggestion for formatting

```suggestion
    "dev:load:commonjs": "export RUNTIME_ENVIRONMENT=development; yarn build && time node -e 'const StrongConfig = require(\".\"); console.log(\"Example Config:\\n\", (new StrongConfig()).load(\"./example\"))'",
```

Before:

image


After:

image

package.json
```diff
@@ -22,7 +22,12 @@
...
+    "dev:load:es6": "export RUNTIME_ENVIRONMENT=development; time ts-node -e 'import StrongConfig from \".\"; console.log(\"Example Config:\", (new StrongConfig()).load(\"example\"))'",
```
chapati23: same formatting suggestion here, and also i _think_ we can shorten the bash command a tiny bit here (you can define env vars in front of any bash command that are valid for this specific command):

```suggestion
    "dev:load:es6": "RUNTIME_ENVIRONMENT=development time ts-node -e 'import StrongConfig from \".\"; console.log(\"Example Config:\\n\", (new StrongConfig()).load(\"example\"))'",
```

nodemon.json
```diff
@@ -0,0 +1,5 @@
+{
+  "ignore": ["**/*.test.ts", ".git", "node_modules"],
```
chapati23: I think we can simplify this to: 

```suggestion
  "ignore": ["**/*.test.ts"],
```

Because `.git` and `node_modules` are ignored per default: https://github.com/remy/nodemon#ignoring-files

src/index.ts
```diff
@@ -16,12 +16,12 @@ export = class StrongConfig {
...
   }
 
-  public load(): ReturnType<typeof load> {
+  public load(configDir?: string): ReturnType<typeof load> {
     if (this.config) {
       return this.config
     }
 
-    this.config = load()
+    this.config = load(configDir)
```
chapati23, line 24: Can this lead to a bug when `configDir` is falsy and we're passing e.g. `null` down to the `load()` function?

chapati23 (changes requested):
great idea with including `time` 👍

just a few small comments

package.json
mohoff: ah double backslash it is... sweet - my `\n` didn't work

package.json
mohoff: Do you know if that works on Windows+Linux? Any ways to tests this without having access to a Windows/Linux machine?

src/index.ts
mohoff: Good point, think we should check for that inside `load()`. Values of `configDir` we want to fail on are `typeof !== 'string'` which also catches `null`. Maybe we can make use of a util `isPath(configDir)` or smth too here.

nodemon.json
mohoff: Jup, thanks

package.json
mohoff: We could even look for a json pretty print util here. Wdyt?

package.json
chapati23: For Linux you can just ssh into any linux-based docker container on your local machine

windows: good point, to be honest i don't know about syntax there. i usually test using one of their free VMs: https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/

(btw this is something where gitlab actions will be nice as they offer different OS runners for testing)

package.json
chapati23: yeah, that'd be even better!

src/index.ts
chapati23: Maybe safer to also preset it to `(configDir = "./config")` in the `StrongConfig.load()` class method?

mohoff: the default param will only overwrite `(not passed)` or `undefined`. `null`, any `number`, etc., would all slip through

package.json
mohoff: done

package.json
mohoff: used `cross-env` now

nodemon.json
mohoff: removed

src/index.ts
mohoff: added `path.normalize` now, which throws when a non-string is passed https://nodejs.org/api/path.html#path_path_normalize_path

Regarding adding a second default parameter:

1. I think it's redundant since we're doing that in `load()` already.
2. I'd leave all things validation/defaulting/etc inside the `load()` function. It keeps `index.ts` simple.

src/load.ts
```diff
@@ -7,12 +9,17 @@ import * as sops from './utils/sops'
 
 import { HydratedConfig } from './types'
 
-export const load = (): HydratedConfig => {
+export const load = (configDir = './config'): HydratedConfig => {
+  const normedConfigDir = path.normalize(configDir)
```
chapati23: nit: i think `normedConfigDir` is incorrect english here (at least i have never seen `normed` as a var name anywhere else so far)

```suggestion
  const normalizedConfigDir = path.normalize(configDir)
```

chapati23 (changes requested):
just one nit left

src/index.ts
chapati23: perfect, sounds good

src/load.ts
mohoff: updated

chapati23 (approved):

PR #23, opened 2019-10-14
chore: add post-merge git hook used by husky

mohoff:
Donezo:

- [x] Add script that is used by husky as post-merge and post-rewrite hook.


chapati23 (approved):
Ha! It was empty! Curious how that happened 🤔

PR #27, opened 2019-11-04
feat: parametrize StrongConfig

### LINKED ISSUE #8: Allow configuration of strong-config by passing a config obj into constructor
The constructor of StrongConfig should accept a configuration object with various fields that configure the behavior of strong-config.

Preliminary configuration fields:

```
runtimeEnvironmentVariable: string (DEFAULT: 'runtimeEnvironment')
generateTypes: boolean (DEFAULT: true)
generatedTypesPath: string (DEFAULT: 'strongconfig.d.ts')
generatedTypeRoot: string   (DEFAULT: 'Config')
substitutionPattern: string   (DEFAULT: /\${(\w+)}/g)
environments: string[] (DEFAULT: ['development', 'production'])
configDir: string (DEFAULT: 'config/')
schemaFile: string (DEFAULT: 'config/schema.json')
logger: Logger (DEFAULT: internal bunyan instance) --> interface matters here: debug/info/warn/error?
```

Things to take care of:

- [ ] Support passing a configuration object
- [ ] Validate the passed configuration object
- [ ] Replace existing hard-coded values in code with configuration obj values

mohoff:
Donezo:

- [x] add `/src/params`, which contains parameters-related stuff.
    * default parameters
    * parameter validation via json-schema
    * parameter types.
- [x] pass around parameters where needed. General philosophy: pass parameter object as a whole so that called function can pick the params it requires via object-destructuring.
- [x] Update/add tests
- [x] Update `yarn dev:*` scripts (basically, now paths go into class instead of methods) 
- [x] Run `yarn dev:*` scripts as part of healthcheck.ts

Misc:
- [x] fixed a type related to `innerHydrateFunction`
- [x] cleaning up variadic arguments in `validate.ts`
- [x] rename `validate-config` to `validate-json` as it's generic and we use it for validating parameters too

How to test:
- [x] Generally, everything should work like before since this is primarily a refactor.
- [x] Run `yarn dev:*` scripts, they still work as intended
- [x] Overwrite parameters and verify the corresponding change actually happens. You can also use the `yarn dev:*` scripts for this.
- [x] Pass invalid parameters --> will be caught by json validation.

fixes https://github.com/strong-config/node/issues/8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,6 @@
 lib
+# Default output for generated types when strongConfig.load() is called
+strong-config.d.ts
 
 # Logs
 logs
--- a/package.json
+++ b/package.json
@@ -23,11 +23,11 @@
     "watch": "tsc --watch",
     "clean": "rimraf build/",
     "build": "tsc",
-    "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env NODE_ENV=development time node scripts/dev/load-commonjs.js | jq .",
-    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env NODE_ENV=development time ts-node --transpile-only scripts/dev/load-es6.ts | jq .",
+    "dev:load:commonjs": "yarn build; cross-env NODE_ENV=development time node scripts/dev/load-commonjs.js",
+    "dev:load:es6": "cross-env NODE_ENV=development time ts-node --transpile-only scripts/dev/load-es6.ts",
     "dev:load": "yarn dev:load:es6",
     "dev:load:watch": "nodemon --exec 'yarn dev:load:es6'",
-    "dev:validate": "echo 'Validation Result:\n'; cross-env NODE_ENV=development time ts-node --transpile-only scripts/dev/validate.ts",
+    "dev:validate": "cross-env NODE_ENV=development time ts-node --transpile-only scripts/dev/validate.ts",
     "dev:validate:watch": "nodemon --exec 'yarn dev:validate'"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -23,11 +23,11 @@
     "watch": "tsc --watch",
     "clean": "rimraf build/",
     "build": "tsc",
-    "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time node -p -e 'const StrongConfig = require(\".\"); JSON.stringify((new StrongConfig()).load(\"./example\"))' | jq .",
-    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T -p -e 'import StrongConfig from \".\"; JSON.stringify((new StrongConfig()).load(\"example\"))' | jq .",
+    "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time node -p -e 'const StrongConfig = require(\"./lib\"); JSON.stringify((new StrongConfig({ configPath: \"example/\", schemaPath:\"example/schema.json\" })).load())' | jq .",
+    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T -p -e 'import StrongConfig from \"./src\"; JSON.stringify((new StrongConfig({ configPath: \"example/\", schemaPath:\"example/schema.json\" })).load())' | jq .",
     "dev:load": "yarn dev:load:es6",
     "dev:load:watch": "nodemon --exec 'yarn dev:load:es6'",
-    "dev:validate": "echo 'Validation Result:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T -p -e 'import StrongConfig from \".\"; (new StrongConfig()).validate(\"example/schema.json\", \"example\")'",
+    "dev:validate": "echo 'Validation Result:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T -p -e 'import StrongConfig from \"./src\"; (new StrongConfig({ configPath: \"example/\", schemaPath:\"example/schema.json\" })).validate()'",
     "dev:validate:watch": "nodemon --exec 'yarn dev:validate'"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "author": "Brickblock Engineering <dev@brickblock.io>",
   "license": "MIT",
   "scripts": {
-    "health": "ts-node scripts/healthcheck.ts",
+    "health": "ts-node -T scripts/healthcheck.ts",
     "lint": "yarn lint:ts && yarn lint:json && yarn lint:markdown",
     "lint:ts": "eslint --ext [.ts] src scripts",
     "lint:ts:fix": "eslint --fix --ext [.ts] src scripts",
@@ -23,11 +23,11 @@
     "watch": "tsc --watch",
     "clean": "rimraf build/",
     "build": "tsc",
-    "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time node -p -e 'const StrongConfig = require(\"./lib\"); JSON.stringify((new StrongConfig({ configPath: \"example/\", schemaPath:\"example/schema.json\" })).load())' | jq .",
-    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T -p -e 'import StrongConfig from \"./src\"; JSON.stringify((new StrongConfig({ configPath: \"example/\", schemaPath:\"example/schema.json\" })).load())' | jq .",
+    "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time node scripts/dev/load-commonjs.js | jq .",
+    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T scripts/dev/load-es6.ts | jq .",
     "dev:load": "yarn dev:load:es6",
     "dev:load:watch": "nodemon --exec 'yarn dev:load:es6'",
-    "dev:validate": "echo 'Validation Result:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T -p -e 'import StrongConfig from \"./src\"; (new StrongConfig({ configPath: \"example/\", schemaPath:\"example/schema.json\" })).validate()'",
+    "dev:validate": "echo 'Validation Result:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T scripts/dev/validate.ts",
     "dev:validate:watch": "nodemon --exec 'yarn dev:validate'"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -23,11 +23,11 @@
     "watch": "tsc --watch",
     "clean": "rimraf build/",
     "build": "tsc",
-    "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time node scripts/dev/load-commonjs.js | jq .",
-    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T scripts/dev/load-es6.ts | jq .",
+    "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env NODE_ENV=development time node scripts/dev/load-commonjs.js | jq .",
+    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env NODE_ENV=development time ts-node -T scripts/dev/load-es6.ts | jq .",
     "dev:load": "yarn dev:load:es6",
     "dev:load:watch": "nodemon --exec 'yarn dev:load:es6'",
-    "dev:validate": "echo 'Validation Result:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T scripts/dev/validate.ts",
+    "dev:validate": "echo 'Validation Result:\n'; cross-env NODE_ENV=development time ts-node -T scripts/dev/validate.ts",
     "dev:validate:watch": "nodemon --exec 'yarn dev:validate'"
   },
   "devDependencies": {
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
   "author": "Brickblock Engineering <dev@brickblock.io>",
   "license": "MIT",
   "scripts": {
-    "health": "ts-node -T scripts/healthcheck.ts",
+    "health": "ts-node --transpile-only scripts/healthcheck.ts",
     "lint": "yarn lint:ts && yarn lint:json && yarn lint:markdown",
     "lint:ts": "eslint --ext [.ts] src scripts",
     "lint:ts:fix": "eslint --fix --ext [.ts] src scripts",
@@ -24,10 +24,10 @@
     "clean": "rimraf build/",
     "build": "tsc",
     "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env NODE_ENV=development time node scripts/dev/load-commonjs.js | jq .",
-    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env NODE_ENV=development time ts-node -T scripts/dev/load-es6.ts | jq .",
+    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env NODE_ENV=development time ts-node --transpile-only scripts/dev/load-es6.ts | jq .",
     "dev:load": "yarn dev:load:es6",
     "dev:load:watch": "nodemon --exec 'yarn dev:load:es6'",
-    "dev:validate": "echo 'Validation Result:\n'; cross-env NODE_ENV=development time ts-node -T scripts/dev/validate.ts",
+    "dev:validate": "echo 'Validation Result:\n'; cross-env NODE_ENV=development time ts-node --transpile-only scripts/dev/validate.ts",
     "dev:validate:watch": "nodemon --exec 'yarn dev:validate'"
   },
   "devDependencies": {
--- a/scripts/dev/load-es6.ts
+++ b/scripts/dev/load-es6.ts
@@ -1,0 +1,10 @@
+import StrongConfig from '../../src'
+
+const strongConfig = new StrongConfig({
+  configPath: 'example/',
+  schemaPath: 'example/schema.json',
+})
+
+const config = strongConfig.load()
+
+console.log(JSON.stringify(config))
--- a/scripts/dev/load-es6.ts
+++ b/scripts/dev/load-es6.ts
@@ -1,3 +1,4 @@
+import { inspect } from 'util'
 import StrongConfig from '../../src'
 
 const strongConfig = new StrongConfig({
@@ -7,4 +8,6 @@ const strongConfig = new StrongConfig({
 
 const config = strongConfig.load()
 
-console.log(JSON.stringify(config))
+console.log('\nLoaded Config:\n')
+console.log(inspect(config, { colors: true, compact: false }))
+console.log('\n')
--- a/src/index.ts
+++ b/src/index.ts
@@ -1,35 +1,41 @@
+import R from 'ramda'
+
 import { load } from './load'
 import { validate } from './validate'
+import { validate as validateParameters } from './params/validate'
 
 import { MemoizedConfig } from './types'
-
-interface Parameters {
-  // TODO: define Parameters interface
-}
+import { defaultParameters, Parameters } from './params'
 
 export = class StrongConfig {
-  public readonly params: Parameters | undefined
+  public readonly parameters: Parameters
   private config: MemoizedConfig
 
-  constructor(params?: Parameters) {
-    // TODO: validate params
-    this.params = params
+  constructor(parameters?: Partial<Parameters>) {
+    this.parameters = validateParameters(
+      parameters
+        ? {
+            ...defaultParameters,
+            ...parameters,
+          }
+        : defaultParameters
+    )
   }
 
-  public load(configDir?: string): ReturnType<typeof load> {
+  public load(): ReturnType<typeof load> {
     if (this.config) {
       return this.config
     }
 
-    this.config = load(configDir)
+    this.config = load(this.parameters)
 
     return this.config
   }
 
-  public validate(
-    schemaPath: string,
-    ...configPaths: string[]
-  ): ReturnType<typeof validate> {
-    return validate(schemaPath, ...configPaths)
+  public validate(...configPaths: string[]): ReturnType<typeof validate> {
+    return validate(
+      R.isEmpty(configPaths) ? [this.parameters.configPath] : configPaths,
+      this.parameters
+    )
   }
 }
--- a/src/index.ts
+++ b/src/index.ts
@@ -2,23 +2,23 @@ import R from 'ramda'
 
 import { load } from './load'
 import { validate } from './validate'
-import { validate as validateParameters } from './params/validate'
+import { validate as validateOptions } from './options/validate'
 
 import { MemoizedConfig } from './types'
-import { defaultParameters, Parameters } from './params'
+import { defaultOptions, Options } from './options'
 
 export = class StrongConfig {
-  public readonly parameters: Parameters
+  public readonly options: Options
   private config: MemoizedConfig
 
-  constructor(parameters?: Partial<Parameters>) {
-    this.parameters = validateParameters(
-      parameters
+  constructor(options?: Partial<Options>) {
+    this.options = validateOptions(
+      options
         ? {
-            ...defaultParameters,
-            ...parameters,
+            ...defaultOptions,
+            ...options,
           }
-        : defaultParameters
+        : defaultOptions
     )
   }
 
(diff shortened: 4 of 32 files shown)


src/params/index.ts
```diff
@@ -0,0 +1,23 @@
...
+  filePath: string
+}
+
+export interface Parameters {
+  runtimeEnvName: string
+  types: TypesParameters
+  substitutionPattern: string
+  configPath: string
+  schemaPath: string
+}
```
mohoff: These types could live in the top-level `types.ts` as `Parameters` is imported in many files. Opinions?

src/validate.ts
```diff
@@ -3,28 +3,24 @@ import fs from 'fs'
...
-  ...configPaths: string[]
+  configPaths: string[],
+  { schemaPath }: Parameters
 ): true => {
   const normalizedSchemaPath = path.normalize(schemaPath)
-
-  if (R.isNil(configPaths) || R.isEmpty(configPaths)) {
-    throw new Error('Config path(s) must be passed so they can be validated')
-  }
-
```
mohoff, line 27: This is not needed anymore:

- It can't be nil as the default config path is coming from `defaultParameters` (which is validated too, so can't be nil)
- It can't be empty (`[]`) either. The default is not empty and an empty array can't be constructed with variadic arguments. No argument to `strongConfig.validate()` would result in the default value being used instead.

package.json
```diff
@@ -23,11 +23,11 @@
     "watch": "tsc --watch",
     "clean": "rimraf build/",
     "build": "tsc",
-    "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time node -p -e 'const StrongConfig = require(\".\"); JSON.stringify((new StrongConfig()).load(\"./example\"))' | jq .",
-    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T -p -e 'import StrongConfig from \".\"; JSON.stringify((new StrongConfig()).load(\"example\"))' | jq .",
+    "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time node -p -e 'const StrongConfig = require(\"./lib\"); JSON.stringify((new StrongConfig({ configPath: \"example/\", schemaPath:\"example/schema.json\" })).load())' | jq .",
```
crosby: Not strictly related to this PR I know, but these commands are pretty hard to parse. What do you think about adding a simple `/scripts` dir with some of these in?

src/index.ts
```diff
@@ -1,35 +1,41 @@
...
 export = class StrongConfig {
-  public readonly params: Parameters | undefined
+  public readonly parameters: Parameters
   private config: MemoizedConfig
 
-  constructor(params?: Parameters) {
-    // TODO: validate params
-    this.params = params
+  constructor(parameters?: Partial<Parameters>) {
+    this.parameters = validateParameters(
```
crosby: Strictly speaking, validation shouldn't occur in the constructor method: 

- http://misko.hevery.com/code-reviewers-guide/flaw-constructor-does-real-work/
- http://blog.millermedeiros.com/constructors-should-not-cause-side-effects/

Though it seems like the obvious place to put it, I would instead recommend validating the values in the methods that require the params e.g. `load()`

src/utils/validate-json.ts
```diff
@@ -1,13 +1,13 @@
 import Ajv from 'ajv'
 
-import { BaseConfig, Schema } from '../types'
+import { JSONObject, Schema } from '../types'
 
 const ajv = new Ajv({ useDefaults: true })
 
-export const validateConfig = (config: BaseConfig, schema: Schema): true => {
+export const validateJson = (jsonObject: JSONObject, schema: Schema): true => {
   const validate = ajv.compile(schema)
```
crosby, line 8: I'd recommend memoizing this function: https://ramdajs.com/docs/#memoizeWith

Every time this `validateJson` is called, it'll recompile the schema which probably isn't necessary.

src/utils/read-file.ts
```diff
@@ -38,24 +38,10 @@ export const readConfigFile = (basePath: string, fileName: string): File => {
...
-export const readSchemaFile = (
-  basePath: string,
-  fileName = 'schema.json'
-): File | undefined => {
-  const filePaths = findFiles(basePath, fileName)
-
-  if (R.isEmpty(filePaths)) {
+export const readSchemaFile = (schemaPath: string): File | undefined => {
+  if (R.isNil(schemaPath) || !isJson(schemaPath)) {
     return undefined
```
crosby: Should this maybe throw an error instead? Missing schema seems important.

src/utils/generate-type-from-schema.ts
```diff
@@ -2,19 +2,27 @@ import { compileFromFile } from 'json-schema-to-typescript'
...
-  filePath: string
-): Promise<void> => {
-  const baseTypes = await compileFromFile(filePath)
+export const generateTypeFromSchema = async ({
+  runtimeEnvName,
+  schemaPath,
+  types,
+}: Parameters): Promise<void> => {
+  if (types === false) {
+    return
```
crosby: In this case, this Promise technically should be rejected. I'd recommend throwing an error here so that the caller of this function can know to do something practical in the case that no types were generated.

src/utils/find-files.ts
```diff
@@ -11,6 +11,8 @@ export const isSchema = (filePath: string): boolean =>
     R.split('/')
   )(filePath)
 
+export const isJson = (filePath: string): boolean => filePath.endsWith('.json')
```
crosby, line 14: IMO this is too short to justify the additional function. You don't really gain anything by doing `isJson(filePath)` instead of `filePath.endsWith('.json')`.

src/params/validate.ts
```diff
@@ -0,0 +1,9 @@
+import { validateJson } from '../utils/validate-json'
+import { parametersSchema } from './schema'
+
+import { JSONObject } from '../types'
+import { Parameters } from '.'
+
+export const validate = (parameters: Parameters): Parameters =>
+  validateJson((parameters as unknown) as JSONObject, parametersSchema) &&
```
crosby: Not familiar with this `as unknown` syntax. What's happening here?

src/params/schema.ts
```diff
@@ -0,0 +1,69 @@
...
+    },
+    types: {
+      title: 'Parameters related to types',
+      description:
+        'Type-related parameters controlling the generation of Typescript types for the config',
+      type: ['object'],
+      additionalProperties: false,
+      properties: {
+        rootTypeName: {
+          title: 'The name of the generated root type',
```
crosby: I think just `Root type name` here is fine. Having the `title` and `description` the same is a bit redundant.

src/params/schema.ts
```diff
@@ -0,0 +1,69 @@
...
+    'runtimeEnvName',
+    'types',
+    'substitutionPattern',
+    'configPath',
+    'schemaPath',
+  ],
+  additionalProperties: false,
+  properties: {
+    runtimeEnvName: {
+      title: 'The name of the runtime environment variable',
```
crosby: Same here.

crosby: Actually, just all titles here can just be much shorter. They all read like descriptons.

src/params/index.ts
```diff
@@ -0,0 +1,23 @@
...
+export interface Parameters {
+  runtimeEnvName: string
+  types: TypesParameters | false
+  substitutionPattern: string
+  configPath: string
+  schemaPath: string
+}
+
+export const defaultParameters: Parameters = {
+  runtimeEnvName: 'RUNTIME_ENVIRONMENT',
```
crosby: Should we consider using `NODE_ENV` here? I think it'll be what a lot of people would default to using.

src/load.test.ts
```diff
@@ -1,18 +1,22 @@
...
-import { validateConfig } from './utils/validate-config'
+import { validateJson } from './utils/validate-json'
 import { readConfigFile, readSchemaFile } from './utils/read-file'
 import { decryptToObject } from './utils/sops'
 
 import { HydratedConfig } from './types'
 
+const mockedParameters = defaultParameters
+// Note: This variable must be named according to defaultParameters.runtimeEnvName
+const RUNTIME_ENVIRONMENT = 'development'
```
crosby: For tests maybe make this `test`?

package.json
mohoff: In thee sense of `.sh` or `.ts` scripts?

src/index.ts
mohoff: Thanks for the links. I still think the constructor is the best place.

The links seem to more refer to _side effects_, which is not what we have. I like the current contract of _if you pass valid params, you can create a strong-config instance. If not, you can't_. That's deterministic and is not a side-effect.

Additionally, validating in the methods mean we duplicate the validation logic in multiple places instead of validation once for one instance (which is sufficient since params are not designed to change at runtime).

src/utils/read-file.ts
mohoff: We support having no schema on`load()`

src/utils/generate-type-from-schema.ts
mohoff: Hm not sure I agree with throwing. Not generating types is supported and not an error case.

Does it make sense to move this `if` up to the caller? So `generateTypeFromSchema(..., types)` assumes valid non-falsy `types`

src/params/validate.ts
mohoff: Would be very curious to know a cleaner way of doing this, too.

It's basically casting from type `Parameters` to `JSONObject`. Typescript doesn't recognize that the former is a specialized type of the latter. But I think that's my incapability of making the type correct, maybe we need an `infer` .. some better type magic for sure.

src/params/index.ts
mohoff: Good point. Having found https://stackoverflow.com/a/16979503, I agree to `NODE_ENV`

src/load.test.ts
mohoff: Then we can't use `defaultParameters` anymore OOTB. No strong opinion

src/params/schema.ts
mohoff: Okay

package.json
crosby: The `const StrongConfig = require(\"./lib\"); JSON.stringify((new StrongConfig({ configPath: \"example/\", schemaPath:\"example/schema.json\" })).load())` part I think ought to just be in it's own file.

Or wouldn't this work by loading using the CLI tool?

src/index.ts
crosby: A "side-effect" in programming refers to any observable effect besides returning a value. Since constructors should not return values they shouldn't return errors either.

What's more, the validation method actually loads a file (the JSON schema) and then validates a JSON object against it. 

The biggest problem with constructors that do too much (doing anything other than assigning params to class variables is too much), is that constructors by their nature cannot be async. 

In JS this is a problem because it's single-threaded. This means that if for some reason, the JSON schema takes a long time to load (which any IO operation can do), then the whole thread is blocked because running `new StrongConfig` *must* block even though you' might've asynced your validation methods.

There are numerous other disadvantages across different languages:

- If an error is thrown in the constructor then object lifecycles can't run destructors so memory leaks occur
- It's much harder to test objects that do work inside the constructor
- It violates developer expectations e.g. you don't expect that just doing `new StrongConfig` would do file IO, JSON parsing, validation etc

Bloated constructors really are an anti-pattern.

src/utils/read-file.ts
crosby: I think `null` might be a more appropriate return here: https://airbrake.io/blog/javascript-error-handling/null-undefined-properties

src/utils/generate-type-from-schema.ts
crosby: I would argue that this _is_ an error case for this function. If `generateTypeFromSchema` cannot generate a type from a schema because it has no types it's an error instead of a success.

crosby: If the calling function doesn't mind that generating the types errored, then the error can be ignored there.

src/load.test.ts
crosby: I generally think it's good to treat testing as it's own environment. For example, Jest sets `NODE_ENV` to `test` when testing: https://stackoverflow.com/questions/48461394/node-env-with-jest

src/index.ts
mohoff: Thanks for the detailed reply.

I agree with
- constructors shouldn't do much
- no async action possible in constructors, single threaded JS, etc.
- no side effects (no disk IO, no http calls, etc.)

but disagree with all the rest though
- we're not really reading a file in the constructor a la `fs.readFile`. I mean it's just another import: `import { parametersSchema } from './schema'`, see `params/validate.ts`.
- Why much harder to test in our case? I test the `validateParameters` isolatedly, and, assert on `strongConfig.parameters` when constructing the class. Straight forward unit tests.
- destructors assume a previously constructed object. If construction fails, there can't be a leak. Lacking in-depth knowledge here but the links below support my guess.
- Nothing is 'bloated' in this constructor. User input validation is not an anti-pattern for me, including constructor logic. 

In fact, throwing inside a constructor seems to be totally fine in programming

- https://stackoverflow.com/questions/6086334/is-it-good-practice-to-make-the-constructor-throw-an-exception/6086399
- https://stackoverflow.com/questions/810839/throwing-exceptions-from-constructors
- https://stackoverflow.com/questions/5149619/what-should-a-javascript-constructor-return-if-it-fails
- https://stackoverflow.com/questions/32323406/what-happens-if-a-constructor-throws-an-exception/32323458

src/utils/read-file.ts
mohoff: aight

src/utils/generate-type-from-schema.ts
mohoff: That's why I suggested to move the if-statement up. So that `generateTypeFromSchema` is only called when we generate types for sure.

.gitignore
```diff
@@ -1,4 +1,6 @@
 lib
+# Default output for generated types when strongConfig.load() is called
```
chapati23, line 2: 👍 for the comment, always better to err on the side of over-documenting in open source projects

package.json
chapati23: Agree with Richard here. Up to a certain point IMO it's ok or even better to just inline the command, but I think we're past the point here and it gets hard to decipher. +1 for extracting these into `/scripts/load.ts` with an argument or `/scripts/load-commonjs.ts` and `/scripts/load-es6.ts`

src/params/index.ts
```diff
@@ -0,0 +1,23 @@
+export interface TypesParameters {
+  rootTypeName: string
+  filePath: string
+}
+
+export interface Parameters {
```
chapati23: A general naming discussion: **Options** or **Parameters**.

From my totally subjective anecdotal experience, it seems most projects call these things **Options**. E.g. many JS libs will accept an `options` argument that is an object of flags or config.

src/params/schema.ts
```diff
@@ -0,0 +1,69 @@
...
+        rootTypeName: {
+          title: 'The name of the generated root type',
+          description: 'The name of the generated root type',
+          examples: ['Config', 'AppConfig'],
+          type: 'string',
+          pattern: '^[A-Z]\\w*$',
+        },
+        filePath: {
+          title: 'The path of the generate type file',
+          description: 'The file that the generated types should be stored to',
```
chapati23: nit: **stored in**

src/utils/generate-type-from-schema.test.ts
```diff
@@ -62,33 +63,47 @@ describe('generateTypeFromSchema()', () => {
     jest.clearAllMocks()
   })
 
+  it('immediately returns when types=false', async () => {
```
chapati23: nit: either spell out the sentence or use real programming syntax `when types === false`

src/utils/generate-type-from-schema.ts
chapati23: Would agree with both here. Inside this function a non-existant schema is an error IMO. So we should ideally also not trying to call it without one.

src/utils/find-files.ts
```diff
@@ -11,6 +11,8 @@ export const isSchema = (filePath: string): boolean =>
...
+
 export const findFiles = (
```
chapati23, line 16: is this function still used anywhere after this PR is merged?

chapati23 (commented):
Great work overall @mohoff. Just some small comments and one big one: There is no "How To Test" in the PR description and I didn't have enough time to guess what or how I should or should not test so all I did was glance over the code for now.

Through your compressed "Donezo" style description, you make it harder to verify that the new code fixes the actual problem and lower the probability that reviewers will actually run the code (vs. just reading through it)

src/utils/validate-json.ts
mohoff: Thanks for the pointer - that's interesting.

I think we could make this work in theory but honestly, I think we don't need it: There are only at most two validations happening (with distinct schemas) in the lifecycle of a `strongConfig` instance:

1. validation of parameters (assuming we continue to do this once at class construction)
2. validation of one config file (optional, users can have no schema file)

I conclude that there are not repeated validations happening where we could benefit from memoized compiled schemas.

src/params/index.ts
mohoff: Had the same thoughts. Yeah, honestly feel the same

src/utils/find-files.ts
mohoff: Just in `findConfigFiles()` below

package.json
mohoff: I like this idea too, don't get me wrong. Outsourcing the ts-part it is 👍

mohoff:
Added a How-to-test section to the PR description. Hope it helps. This is primarily a refactor so the spirit is "Everything should work as before".

src/load.test.ts
mohoff: Convinced, will update

package.json
mohoff: updated

src/utils/read-file.ts
mohoff: updated 👍

src/utils/generate-type-from-schema.ts
mohoff: updated as discussed 👍

src/params/schema.ts
mohoff: done

src/params/schema.ts
mohoff: done

src/params/index.ts
mohoff: updated

src/load.test.ts
mohoff: done

src/params/index.ts
mohoff: renamed

src/params/schema.ts
mohoff: thx

src/utils/generate-type-from-schema.test.ts
mohoff: this test is removed. But paid attention to that in the added tests to `load.test.ts`

mohoff:
Note: Added another small improvement - `yarn dev:*` scripts are run in `yarn health` as well now.

package.json
```diff
@@ -11,7 +11,7 @@
   "author": "Brickblock Engineering <dev@brickblock.io>",
   "license": "MIT",
   "scripts": {
-    "health": "ts-node scripts/healthcheck.ts",
+    "health": "ts-node -T scripts/healthcheck.ts",
```
chapati23: nit: generally i'd default to using the long-form arg names because they're easier to understand

```suggestion
    "health": "ts-node --transpile-only scripts/healthcheck.ts",
```

package.json
```diff
@@ -23,11 +23,11 @@
     "watch": "tsc --watch",
     "clean": "rimraf build/",
     "build": "tsc",
-    "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time node -p -e 'const StrongConfig = require(\".\"); JSON.stringify((new StrongConfig()).load(\"./example\"))' | jq .",
-    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env RUNTIME_ENVIRONMENT=development time ts-node -T -p -e 'import StrongConfig from \".\"; JSON.stringify((new StrongConfig()).load(\"example\"))' | jq .",
+    "dev:load:commonjs": "yarn build; echo 'Loaded Config:\n'; cross-env NODE_ENV=development time node scripts/dev/load-commonjs.js | jq .",
+    "dev:load:es6": "echo 'Loaded Config:\n'; cross-env NODE_ENV=development time ts-node -T scripts/dev/load-es6.ts | jq .",
```
chapati23: Small improvement suggestion to get rid of the `echo` and the `jq` dependency here: (check the corresponding suggestion in `scripts/load-es6.ts`

```suggestion
    "dev:load:es6": "cross-env NODE_ENV=development time ts-node --transpile-only scripts/dev/load-es6.ts",
```

Same comment would apply to the other yarn tasks here, just wanted to align on the general approach first

scripts/dev/load-es6.ts
```diff
@@ -0,0 +1,10 @@
+import StrongConfig from '../../src'
+
+const strongConfig = new StrongConfig({
+  configPath: 'example/',
+  schemaPath: 'example/schema.json',
+})
+
+const config = strongConfig.load()
+
+console.log(JSON.stringify(config))
```
chapati23:
```suggestion
console.log('\nLoaded Config:\n')
console.dir(config, { colors: true })
console.log('\n')
```

chapati23 (changes requested):
good, i think we're almost there, just some small suggestions on improving the dev scripts

src/validate.ts
chapati23: thanks for the explainer, that was helpful to me

package.json
mohoff: I tested this but it seems to show a general `console.log` type of output, which is why I went for `| jq .`. Can't see a difference with `console.dir` and `{colors: true}`. `jq` does a much better job in color-coding and indentation.

Also, `jq` would be just a dependency for developers, imo acceptable

package.json
mohoff: ok, updated in next git push

src/utils/validate-json.ts
crosby: That's a fair point. Nevertheless, I think it'd be good practice to not assume how we might use this `validateJson` function in the future. The resulting code would be:

```
R.memoizeWith(R.identiity, schema => ajv.compile(schema))
```

Now `validateJson` can be called numerous times without any performance penalities due to schema compilation.

It's minor, but perhaps the tests (which do run this multiple times) would be quicker too :)

src/params/index.ts
crosby: IMO these are fine where they are @mohoff

src/index.ts
crosby: After doing a lot more reading on the topic it definitely seems like there's two schools of thought with regards to the right way to do this. Pros and cons to both it looks like.

In this particular case, I would say that you're right. But to address your points:

- I believe that JS imports are file IO...but I can't find anything good here to back up my claim
- Testing difficulty comes when we introduce [subclassing](https://www.oreilly.com/library/view/learning-javascript-design/9781449334840/ch09s12.html)

An example:

```
class A {
  constructor() {
    someFuncThatErrors()
  }
}

class B extends A {
  constructor() {
  }
}
```

When testing A you can simply mock behaviour from `someFuncThatErrors` and successfully assert. When testing B you can't so easily (though I believe it's not too bad in JS). 

- For your point with regards to destructors, memory is allocated *before* object instantiation. If destructors aren't called, this memory is sometimes not cleaned up. I'm also lacking some deep JS knowledge here to know if this is the case, but this is a key point in other languages.
- You're right about user input validation, but I'm sure you'll agree how user input validation can go too far:

```
class User {
  constructor(userId) {
     checkUserExistsInDB(userId)
  }
} 
```

Technically, user input validation (we don't want to construct objects with invalid user IDs). 

Basically what I'm saying is, that for now, your approach is fine, but this pattern can very quickly turn messy and sometimes it's better to just allow objects to instantiate and defer work until later.

Happy to resolve for now, once we've concluded the discussion :)

src/utils/validate-json.ts
mohoff: I think `ajv` is caching schemas by default: https://github.com/epoberezkin/ajv/blob/master/lib/ajv.js - just found that while implementing the `R.memoizeWith`

src/index.ts
mohoff: Thanks again for the details, much appreciated! And yes, I agree.

If we just had to choose between `validating in the constructor` and `validate in the method` without anything else to consider, i'd go for the method, fully agree. My major point in our case though is DRY. Think it's much nicer to validate once where you only need to validate once.

package.json
chapati23: interesting. how did you test this? here's a comparison screenshot from iTerm with my suggestion solution on the left, and the current jq solution on the right:

Screenshot 2019-11-07 at 16 52 30

There's another alternative to `console.dir()` as well:

```js
import { inspect } from 'util'
console.log(inspect(config, { colors: true, compact: false }))
```

which looks like this (again new solution left, old one right)

Screenshot 2019-11-07 at 16 58 28

src/utils/validate-json.ts
crosby: Didn't find where it was caching, can you be more specific?

mohoff: Sorry. `compile()` -> `_addSchema()` --> https://github.com/epoberezkin/ajv/blob/master/lib/ajv.js#L296

package.json
mohoff: alright, `inspect` then. I would have preferred 4 chars much more: `| jq .`

It's updated.

chapati23: thanks, IMO requiring an additional dependency just for the blue color is worse for users in the end. plus more to write in the README ;)

chapati23 (approved):
🎉

PR #29, opened 2019-11-05
chore: fill README

mohoff:
Donezo:

- [x] Write README

fixes #14

Issue #26, opened 2019-10-15
Add 'yarn dev:encrypt' and 'yarn dev:decrypt' tasks


chapati23:
(depends on #25 )

To facilitate local development of the encryption and decryption functionality, we should add yarn tasks to encrypt and decrypt configuration files located in the `example` folder.

To do so, we want to add the following yarn scripts that complement the existing `yarn dev:load` and `yarn dev:validate`:

- `yarn dev:encrypt`: Encrypt the example config files based on a KMS setup that was parametrized while constructing `new StrongConfig(params)`. To make this work, we need an actual KMS setup for testing purposes.
- `yarn dev:decrypt`: Decrypt encrypted example config files. This does not require external parameters, as the `sops`-metadata, which is included in encrypted config files, is sufficient to decrypt the config.

As both are dev commands, we do not require a dynamic runtime environment for now. Instead, we statically encrypt/decrypt, e.g., `/example/development.yml`.

Thus, the scope of this task is:

- [ ] Change example config(s) to contain encrypted fields
- [ ] Implement `yarn dev:encrypt`
- [ ] Implement `yarn dev:decrypt`

mohoff:
What is the desired behaviour of both commands? Do both write the input file in-place or output to a new file? Maybe we want to make syntax as

```
yarn dev:decrypt > decrypted.yaml
```

possible. Thoughts?

chapati23:
I'd say let's start with the current behavior => write to the actual file and also print the results out on the CLI, e.g.:

```sh
NODE_ENV=development yarn dev:encrypt
[output of 'echo config/development.yml']
```

PR #30, opened 2019-11-15
chore(example project): use partially encrypted config with PGP setup

mohoff:
Donezo:

- [x] Use partially encrypted example config instead of no encryption at all
- [x] Provide example PGP keypair (freshly generated) to perform the encryption<->decryption

How to test:

1. Install `gpg`, e.g. with `brew install gpg`
2. Import PGP key via `gpg --import example/pgp/example-keypair.pgp`
3. Now, `yarn dev:load` should just work and output the decrypted config value

This should be mentioned in the README. Depending on which PR goes in first (#29 or this here), we should update the PR going second accordingly.

Facilitates #26


chapati23:
Good idea, just tried it locally and it works as advertised 👍

Will leave the PR open for now because, as you mentioned, the README needs to be slightly amended to explain this.

But code LGTM

mohoff:
We should also make this part of a second, alternative, or extended How-to. Our How-to, which is being created in #29 , explains how to get started with configs without secrets.

This elegance of this PGP approach for a How-to is that we do not need a public GCloud-KMS setup for this. We can demonstrate secrets with a checked-in PGP keypair.

chapati23:
ok, will merge it then!

PR #31, opened 2019-11-19
feat(example project): add dev:encrypt and dev:decrypt

### LINKED ISSUE #26: Add 'yarn dev:encrypt' and 'yarn dev:decrypt' tasks

mohoff:
Donezo:

- [x] Add `yarn dev:[encrypt,decrypt]` which en/decrypt `example/development.yaml` in-place by using the example PGP keypair.

fixes #26

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,8 @@
     "watch": "tsc --watch",
     "clean": "rimraf build/",
     "build": "tsc",
+    "dev:encrypt": "gpg --import --quiet example/pgp/example-keypair.pgp && sops --pgp 2E9644A658379349EFB77E895351CE7FC0AC6E94 --encrypted-suffix Secret --encrypt --in-place example/development.yaml && cat example/development.yaml",
+    "dev:decrypt": "sops --decrypt --in-place example/development.yaml && cat example/development.yaml",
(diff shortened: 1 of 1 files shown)


chapati23 (approved):
LGTM, ran it locally and worked as expected

PR #33, opened 2019-11-21
feat: add CLI with encrypt and decrypt commands

mohoff:
Donezo:

- [x] Add `oclif`-based CLI with `encrypt`, `decrypt` and `validate` commands including required arguments and flags
- [ ] Support all major key providers for encryption.
- [x] Tests. These are finicky: Works when running isolatedly (per-file), but fail when running with `yarn test`. My next attempt is to go a more jest-native approach without `@oclif/test` and with spies.
    - [x] Tests for `decrypt`
    - [x] Tests for `validate`
    - [x] Tests for `encrypt`

How to test:
- Install strong config locally or globally. This should make the CLI available (`strong-config` or `yarn strong-config`)
- Run `strong-config decrypt --help`, `strong-config encrypt --help` and `strong-config validate --help` and explore the CLI. Examples and usage are part of the `--help` output. You can pass the example configs and schemas under `example/`


chapati23:
@mohoff I'm missing some instructions on how to test whether it all works as expected

mohoff:
@chapati23 Added. It's pretty self-documenting. I can add more examples if desired.

Please note that this is WIP/DRAFT

chapati23:
Something I noticed is that there's a noticeable lag when running any `strong-config` command:
image

I don't know it this is a general `@oclif` performance issue or something strong-config specific. But compared to some other local CLIs I use it seems to be slower than it should be, e.g. compare `time tsc --help`:

image

mohoff:
Hm yeah that's a bummer. Will research and experiment (e.g, replace `ts-node` with `node` and work with the transpiled output) soon.

chapati23:
The error messages for `strong-config decrypt a-non-existing-file` and `strong-config decrypt a-folder` and `strong-config decrypt an-existing-js-file.js` could be more helpful:

image

image

image

chapati23:
Trying to re-encrypt an already encrypted file could also use a more instructive error message (e.g. sth like "Looks like <file.yml> is already encrypted 🤔"

image

chapati23:
Maybe we should generally have a pre-check that ensures the `decrypt` and `encrypt` functions only try their magic on proper yaml files, e.g. running this on JS files also just yields cryptic errors:

image

bin/run
```diff
@@ -0,0 +1,15 @@
+#!/usr/bin/env node
+
+/* eslint-disable @typescript-eslint/no-var-requires */
+const fs = require('fs')
+const path = require('path')
+const project = path.join(__dirname, '../tsconfig.json')
```
chapati23, line 6: not sure i fully understand how this works. whenever there exists a `../tsconfig.json` we are in dev mode? are you 100% sure this invariant will always hold?

from the perspective of a consuming app, this would mean that if `./node_modules/@strong-config/node/tsconfig.json` exists, we are in dev mode, correct?

need to make really sure, that `tsconfig.json` never gets bundled with our final prod-bundle then

chapati23: i just remembered something from `bbk-devtools`, that dev mode or no dev mode in oclif clis is determined by a manifest file:

image

i'm struggling to find more info on what it actually does, here's the command: https://www.npmjs.com/package/@oclif/dev-cli#oclif-dev-manifest-path

but if you look at the oclif examples, this file is always mentioned: https://github.com/oclif/example-multi-ts/blob/8edfe090aef770af8f4123cb23e3f1ad63962686/package.json#L39

package.json
```diff
@@ -5,9 +5,27 @@
...
+  "bin": {
+    "strong-config": "./bin/run"
+  },
+  "oclif": {
+    "commands": "./lib/cli/commands",
+    "bin": "strong-config",
+    "plugins": [
+      "@oclif/plugin-help",
+      "@oclif/plugin-warn-if-update-available",
+      "@oclif/plugin-not-found"
```
chapati23: did you have a look at the [autoupdater plugin](https://oclif.io/docs/releasing.html#autoupdater)? could be useful

chapati23: Could the [autocomplete plugin](https://github.com/oclif/plugin-autocomplete) be useful to us?

src/cli/commands/decrypt.test.ts
```diff
@@ -0,0 +1,150 @@
...
+          expect(ctx.stderr).toContain('Decrypting... done')
+          expect(ctx.stderr).toContain('Validating... done')
+        })
+    })
+
+    describe('fails when...', () => {
+      test
+        .command(['decrypt'])
+        .exit(2)
+        .it('no arguments are passed')
```
chapati23: maybe i misunderstand the oclif testing syntax, but this looks like it's not really testing anything to me

chapati23: guessing this is just not implemented yet because PR is still in "draft", if that's the case feel free to resolve this thread @mohoff

chapati23 (changes requested):
Have done a full first pass now, overall I think it's going in the right direction. Summarizing the biggest improvements still needed from my perspective:

1. Handle more unhappy paths (e.g. user passing in bad input such as files in wrong format, or trying to decrypt an already decrypted file etc)
2. Improve performance
3, Support at least one major KMS provider (as already outlined in PR description)

src/cli/commands/decrypt.test.ts
mohoff: Generally, this is the way `@oclif/test` is used (https://oclif.io/docs/testing). In this case I wanted to test for exit code 2. IIRC this works, replacing it with `exit(0)` fails the test.

More importantly though there is more stuff broken with the tests. They work on a per-file basis but running the test suits fails. I think I found an alternative testing way though. more soon

package.json
mohoff: Thanks for the pointer. Will look into it. (resolving the duplicate comment)

mohoff:
> Maybe we should generally have a pre-check that ensures the `decrypt` and `encrypt` functions only try their magic on proper yaml files, e.g. running this on JS files also just yields cryptic errors:
> 
> image

IMO stack traces/cryptic errors aren't so bad. Developers, who will use this tool, can either try to solve the problem themselves with the stack trace or ideally report it as issue on GitHub to help us find the issue. We will have no chance to bugfix or respond to any bugs without error details.

Agreeing with the general idea though. I believe the best way is to have `--verbose/-v` that prints the stack trace. By default, no stack trace but nice error message

bin/run
mohoff: I have to investigate this. I found this file somewhere online and it worked so I was happy.

mohoff:
Getting rid of stack traces is not supported by oclif. There's an open PR though https://github.com/oclif/errors/pull/19 . So I will build something more custom and ditch the oclif-helpers which have proven to be very badly documented and buggy/not working OOTB.

Also, I will ditch `cli-ux` and replace it with `ora` and other tooling i know how to use and know that it is working

mohoff:
The verbose flag is implemented now @chapati23 . Feel free to test it (`--verbose/-v`)

mohoff:
> Maybe we should generally have a pre-check that ensures the `decrypt` and `encrypt` functions only try their magic on proper yaml files, e.g. running this on JS files also just yields cryptic errors:
> 
> image

It should work on json files. Based on your screenshot, `./index.ts` does not exist in your cwd. Try `src/index.ts`.

We should check for 'file does not exist' though

bin/run
mohoff: I don't see `tsconfig.json` in the output dir so seems we can use that.

Thanks for the pointer. It's hard to get more docs about this `oclif-dev` package. I don't see any benefit from using that or a manifest file for now. Tried it and there was no change. Also couldn't find a hint in `bbk-devtools`.

We can use this tool to pack the CLI tool into releases for different platforms though.

mohoff:
> Maybe we should generally have a pre-check that ensures the `decrypt` and `encrypt` functions only try their magic on proper yaml files, e.g. running this on JS files also just yields cryptic errors:
> 
> image

The new `--verbose` includes an error message that says that if found a top-level `sops` field and that the file probably is already decrypted. I'd say this is sufficient. If we do a manual check before running sops (basically just for nicer terminal output), we have to read the file an additional time which comes with performance penalties

src/cli/commands/decrypt.test.ts
chapati23: got it, thx for the link

package.json
chapati23: cool, will leave this open until you made a decision on whether we should use it or not use it

chapati23:
> IMO stack traces/cryptic errors aren't so bad. Developers, who will use this tool, can either try to solve the problem themselves with the stack trace or ideally report it as issue on GitHub to help us find the issue. We will have no chance to bugfix or respond to any bugs without error details.
> 
> Agreeing with the general idea though. I believe the best way is to have `--verbose/-v` that prints the stack trace. By default, no stack trace but nice error message

Yes, I like that idea: Nice error message in plain english per default, stack trace with `--verbose/-v` flag 👍

bin/run
chapati23: ok, it's not a blocker anyways. would just like to get to the bottom of it because i vaguely remember Barath explaining this to me once.

@barathvk do you remember why we needed the `oclif.manifest.json` file in `bbk-devtools`? i also can't seem to find any clear documentation on the oclif pages around it

chapati23:
> Getting rid of stack traces is not supported by oclif. There's an open PR though [oclif/errors#19](https://github.com/oclif/errors/pull/19) . So I will build something more custom and ditch the oclif-helpers which have proven to be very badly documented and buggy/not working OOTB.
> 
> Also, I will ditch `cli-ux` and replace it with `ora` and other tooling i know how to use and know that it is working

sidenote: the github commenting experience on normal PR comments is a bit underwhelming...

to your point: `ora` => 👍, replacing `oclif-helpers` i'm lacking experience with them but i'll trust you'll make the right call here

src/cli/commands/decrypt.ts
```diff
@@ -0,0 +1,86 @@
...
+  static description = 'decrypt config files'
+
+  static strict = true
+
+  static flags = {
+    help: flags.help({
+      char: 'h',
+      description: 'show help',
+    }),
+    verbose: flags.boolean({
```
chapati23, line 24: unfortunately doesn't work for me for some reason. i should have the latest and greatest, just ran:

1. `git pull`
1. `yarn`
1. `yarn build`
1. `strong-config decrypt -v blabla` and a few other syntax variations

No luck:

image

mohoff: forgot to push that. Also investigated the verbose flag in the form `-v/-vv/-vvv` (increasing log verbosity) but it turned out that it isn't supported by oclif at the moment (https://github.com/oclif/oclif/issues/193)

chapati23 (approved):

PR #35, opened 2020-01-08
fix(validation): make schema validation disabled by default and expl…

mohoff:
Donezo:

- [x] Make schema validation on `.load()` disabled per default. Instead, it's explicit opt-in now throught setting `options = {..., schemaPath: 'path-to-schema.json' }`.
- [x] updated options and options' schema
- [x] updated dev scripts
- [x] updated tests

How to test:
- everything should pretty much work as before work as before
- try omitting `schemaPath` / setting it to null / setting it to an invalid path (e.g. in the dev scripts), it should work as expected.

### DIFF
--- a/src/load.ts
+++ b/src/load.ts
@@ -12,7 +12,6 @@ import { Options } from './options'
 
 export const load = (options: Options): HydratedConfig => {
   const normalizedConfigPath = path.normalize(options.configPath)
-  const normalizedSchemaPath = path.normalize(options.schemaPath)
   const runtimeEnv = process.env[options.runtimeEnvName]
 
   if (R.isNil(runtimeEnv)) {
@@ -28,13 +27,20 @@ export const load = (options: Options): HydratedConfig => {
 
   const config = hydrateConfig(runtimeEnv, options)(decrypted)
 
-  const schemaFile = readSchemaFile(normalizedSchemaPath)
+  if (options.schemaPath) {
+    const normalizedSchemaPath = path.normalize(options.schemaPath)
+    const schemaFile = readSchemaFile(normalizedSchemaPath)
 
-  if (schemaFile !== null) {
+    if (schemaFile === null) {
+      throw new Error(
+        `Specified schema at '${options.schemaPath}' is not valid JSON or cannot be read`
+      )
+    }
+
     validateJson(config, schemaFile.contents)
 
     if (options.types !== false) {
-      generateTypeFromSchema(options)
+      generateTypeFromSchema(options.schemaPath, options.types)
     }
   }
 
--- a/src/options/index.ts
+++ b/src/options/index.ts
@@ -8,7 +8,7 @@ export interface Options {
   types: TypeOptions | false
   substitutionPattern: string
   configPath: string
-  schemaPath: string
+  schemaPath: string | null
 }
 
 export const defaultOptions: Options = {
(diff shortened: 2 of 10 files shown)


src/options/index.ts
```diff
@@ -8,7 +8,7 @@ export interface Options {
   types: TypeOptions | false
   substitutionPattern: string
   configPath: string
-  schemaPath: string
+  schemaPath: string | null
```
chapati23, line 11: I've wondered about this actually. What's the best practice in TypeScript here to deal with `null` AND `undefined` at the same time?

I'm talking about the equivalent of a [Maybe type](https://flow.org/en/docs/types/maybe/) in flow: `schemaPath: ?string`

src/options/schema.ts
```diff
@@ -62,8 +62,9 @@ export const optionsSchema = {
         'config/schema.json',
         '../schema.json',
         '/app/config/schema.json',
+        null,
       ],
-      type: 'string',
+      type: ['string', 'null'],
```
chapati23, line 67: Similar question here: Does this include `undefined`?

src/load.ts
```diff
@@ -28,13 +27,20 @@ export const load = (options: Options): HydratedConfig => {
...
   const config = hydrateConfig(runtimeEnv, options)(decrypted)
 
-  const schemaFile = readSchemaFile(normalizedSchemaPath)
+  if (options.schemaPath) {
+    const normalizedSchemaPath = path.normalize(options.schemaPath)
+    const schemaFile = readSchemaFile(normalizedSchemaPath)
+
+    if (schemaFile === null) {
+      throw new Error(
+        'Specified schema at schemaPath is not valid JSON or cannot be read'
```
chapati23: Small improvement suggestion:

```suggestion
        `Specified schema at '${options.schemaPath}' is not valid JSON or cannot be read`
```

src/validate.ts
```diff
@@ -19,6 +19,10 @@ export const validate = (
   configPaths: string[],
   { schemaPath }: Options
 ): true => {
+  if (schemaPath === null) {
```
chapati23, line 22: What if `schemaPath` were `undefined` instead of `null`?

src/options/index.ts
mohoff: typescript is a bit more explicit here, it's `string | null | undefined`. IIRC in function params, `?` notation can be used to mark params as optional though.

src/options/schema.ts
mohoff: no. `undefined` is not valid JSON so not covered by `json-schema`

src/validate.ts
mohoff: i'll play around with this a bit more but i'd say that json-schema catches this so that `schemaPath` can only be a string or `null`.

src/options/schema.ts
chapati23: interesting, didn't know that actually, TIL 💡

src/validate.ts
chapati23: ok, could be right. still seems safer to just make this a standard JS falsy check:

```suggestion
  if (!schemaPath) {
```

somebody might make `schemaPath` just an empty string in the schema, which would still be valid JSON but could cause issues here

mohoff: empty string is basically like path to non-existing file. Both is caught few lines below already.

PR #45, opened 2020-12-23
Improve 'strong-config check' to also detect deeply nested secrets

### LINKED ISSUE #43: strong-config check not recognising secrets in yml object structure
Hi,

I am testing strong-config in a test project of mine but I have problems with `strong-config check`:
Strong-config recognises the following code as having a not encrypted secret:
```yml
api.url: https://test.com
api.apiSecret: top-secret
```
This config is not recognised:
```yml
api:
  url: https://test.com
  apiSecret: top-secret
```

Am I missing something or is this a bug?

chapati23:
### Synopsis
fixes #43 

The first implementation of `strong-config check` was a bit naive as it only looked in the top-level of a potentially deeply-nested object.

This PR adds recursion to the search for secrets and will detect secrets at any object-level.

### How to test
1. Check out the `master` branch first
2. Add a new config file `./example/temp.yaml` with the following contents: 
   ```yaml
   application:
     url: https://myhost.com:8443/myapp/
     username: myname
     very:
       deeply:
         nestedSecret: 123
   ```
3. In the local repo's root, run `./bin/run check example/temp.yaml`
4. See that the check passes **although the file does contain a secret** 🚨🚨🚨
5. Now switch to this branch `git co fix/detect-nested-secrets` (you may need to stash your changes first and then re-apply after switching branches)
6. Run the same check again (`./bin/run check example/temp.yaml`)
7. See that this time it fails (as it should, because there is an unencrypted secret in this file) ✅

### DIFF
--- a/example/unencrypted-with-nested-secrets.yml
+++ b/example/unencrypted-with-nested-secrets.yml
@@ -1,0 +1,6 @@
+application:
+  url: https://myhost.com:8443/myapp/
+  username: myname
+  very:
+    deeply:
+      nestedSecret: 123
--- a/package.json
+++ b/package.json
@@ -85,7 +85,7 @@
   },
   "devDependencies": {
     "@brickblock/eslint-config-base-ts": "^1.0.4",
-    "@commitlint/config-conventional": "^8.3.4",
+    "@commitlint/config-conventional": "^11.0.0",
     "@oclif/dev-cli": "^1.22.2",
     "@types/debug": "^4.1.5",
     "@types/fs-extra": "^9.0.1",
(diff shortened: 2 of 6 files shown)


seantking (approved):

seantking:
Nice 👍 I was wondering if we could use Ramda here in some way instead of the recursive function but I couldn't come up with anything.

chapati23:
yeah, i briefly scrolled through the ramda utils but was too intimidated seeing the many (hundreds?) of puzzle pieces. if you think of a ramda-way to do it, happy for suggestions.

PR #68, opened 2023-03-08
Add ajv formats

radusecrieru:


chapati23:
hey radu, thanks for the PR!
could you give a brief description what the purpose of adding this package is? it doesn't look like it's actively maintained (last commit was in 2021..)

and secondly, if you could add a test that would be appreciated (think the CI check that's failing is because of dropped test coverage)

radusecrieru:
Hey, sorry for the delay 😅 
I was working on an old project that was using the `format: ` structure to validate whether or not the property was a valid URL, and from what I could find at that moment, the ajv formats seemed to add this functionality from their documentation: https://ajv.js.org/guide/formats.html

Since them however, I found that it was an old project, so I deprecated it instead of using the formats.
This change request can be closed.
Thank you!

chapati23:
alright, thanks for double-checking!

